PR #5, opened 2018-07-18
test launch files and add missing dependencies 

jschleicher:
The launch file now starts without errors and has all its dependencies declared:
`roslaunch prbt_moveit_config moveit_planning_execution.launch sim:=false pipeline:=ompl has_gripper:=false`


agutenkunst (approved):
Looks good to me.

PR #6, opened 2018-07-18
Add <url> tag to all package.xml files

agutenkunst:
Since the wiki pages are updated from the package.xml tags:

  - <url type="website">
  - <url type="bugtracker">
  - <url type="repository">

they should exist.

Seems also to be used in the links inside
http://repositories.ros.org/status_page/


jschleicher (approved):
+1

PR #9, opened 2018-07-21
Avoid constant collisions by namespacing them

gavanderhoorn:
As per subject.

Defining constants inside macros automatically namespaces them, which avoids collisions with other xacro macros that happen to define constants with the same name (and `Ln` is rather generic, so the chance of that happening is rather high).


agutenkunst (approved):
I guess here it was overlooked that since Jade local properties can be declared like this.

(http://wiki.ros.org/xacro#Local_properties)

martiniil (approved):
I have no concerns. Approved.

PR #12, opened 2018-08-14
Seperate dependency from gripper.

jschleicher:
The dependency from the gripper was removed to allow easier
transition between ros distros.

See new package [prbt_grippers](https://github.com/PilzDE/prbt_grippers).


agutenkunst (approved):
Looks good

PR #17, opened 2018-09-07
Enable our model to be used in combination with Gazebo

hslusarek:


prbt_support/urdf/prbt.xacro
```diff
@@ -37,6 +37,12 @@ limitations under the License.
     <xacro:include filename="$(find prbt_${gripper}_support)/config/${gripper}_tcp_offset.xacro" />
   </xacro:if>
 
+  <link name="world"/>
+  <joint name="fixed" type="fixed">
```
Chamango90: I would suggest a more descriptive name e.g. `world-base_link`

prbt_support/urdf/prbt.xacro
```diff
@@ -53,4 +59,11 @@ limitations under the License.
     <xacro:include filename="$(find prbt_${gripper}_support)/urdf/${gripper}.urdf.xacro" />
   </xacro:if>
 
+
+  <gazebo>
```
Chamango90: This `gazebo` code, I would either move inside the `prbt_macro.xacro` or together with the `transmission`s in a second dedicated Gazebo xacro (e.g. https://github.com/rosin-project/automatica18_scan_and_plan_demo/blob/c1420a017c6d303fd081e6ce03caba5192884f28/snp_prbt_description/urdf/prbt_gazebo.xacro).

Otherwise every application that uses the `prbt_macro.xacro` with Gazebo would have to add this `gazebo` code by itself.

Chamango90:
Thank you for the PR. I will close mine

prbt_support/urdf/prbt_macro.xacro
```diff
@@ -33,6 +33,10 @@ limitations under the License.
 
     <!-- robot foot link -->
     <link name="${prefix}base_link">
+        <inertial>
+          <mass value="2.4"/>
+          <inertia ixx="1e-3"  ixy="0"  ixz="0" iyy="1e-3" iyz="0" izz="1e-3" />
```
hslusarek: Why not set inertia ixx, iyy, izz to zero?

Chamango90: Even so `base_link` does not move (for most cases) it is not wrong to have an inertia. Probably the only case where this would matter is if an user adds another moving joint before the arm.

agutenkunst (approved):
Accepted

PR #19, opened 2018-09-12
Fix/inertia in cms

jschleicher:
move interial tensors into center-of-mass system
    
* specify center of mass for the links
* reduce inertial tensor accordingly
* unrelated: fix rviz config of test_urdf.launch

### DIFF
--- a/prbt_support/launch/robotmodel.rviz
+++ b/prbt_support/launch/robotmodel.rviz
@@ -39,32 +39,16 @@ Visualization Manager:
         Expand Link Details: false
         Expand Tree: false
         Link Tree Style: Links in Alphabetic Order
+        prbt_base:
+          Alpha: 1
+          Show Axes: false
+          Show Trail: false
+        prbt_base_link:
+          Alpha: 1
+          Show Axes: false
+          Show Trail: false
+          Value: true
         prbt_flange:
-          Alpha: 1
-          Show Axes: false
-          Show Trail: false
-          Value: true
-        prbt_gripper_finger_left_link:
-          Alpha: 1
-          Show Axes: false
-          Show Trail: false
-          Value: true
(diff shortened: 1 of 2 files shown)


prbt_support/launch/robotmodel.rviz
```diff
@@ -39,32 +39,16 @@ Visualization Manager:
...
-          Show Axes: false
-          Show Trail: false
-          Value: true
-        prbt_gripper_finger_left_link_bracket:
+        prbt_base:
           Alpha: 1
           Show Axes: false
           Show Trail: false
-          Value: true
-        prbt_gripper_finger_right_link:
```
hslusarek, line 57: Why do we drop them? Because they are now part of a different package and have no relevance here?

hslusarek:
"unrelated: fix rviz config of test_urdf.launch":
Maybe for clarity we should separate things which are not related to a PullRequest in the future.

prbt_support/urdf/prbt_macro.xacro
```diff
@@ -229,7 +233,8 @@ limitations under the License.
     <link name="${prefix}link_5">
```
hslusarek, line 233: What about the flange? It is still in the representation in which we state the inertia with regard to the link frame. Is this different representation ok?

hslusarek (changes requested):
See comments.

prbt_support/launch/robotmodel.rviz
jschleicher: exactly. Trying to launch `test_urdf.launch` from kinetic-devel failed.

jschleicher:
I've set `origin` of the flange to zero explicitly now and rebased onto kinetic-devel.
@hslusarek please merge before the next binary release. Thanks.

hslusarek (approved):

PR #27, opened 2018-11-07
<gazebo> tag needs to be outside macro

hsd-dev:
The robot does not show up in Gazebo if the <gazebo> tag is inside the macro.

### DIFF
--- a/prbt_support/urdf/prbt_macro.xacro
+++ b/prbt_support/urdf/prbt_macro.xacro
@@ -368,13 +368,12 @@ limitations under the License.
       </joint>
       <actuator name="${prefix}joint_6_motor"/>
     </transmission>
-
-    <gazebo>
-      <plugin name="gazebo_ros_control" filename="libgazebo_ros_control.so" >
-        <robotNamespace>prbt</robotNamespace>
-      </plugin>
-    </gazebo>
-
   </xacro:macro>
+
+  <gazebo>
+    <plugin name="gazebo_ros_control" filename="libgazebo_ros_control.so" >
+      <robotNamespace>prbt</robotNamespace>
+    </plugin>
+  </gazebo>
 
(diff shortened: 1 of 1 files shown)


martiniil:
@ipa-hsd Thank you for contributing.
At the moment we do not consider integration in gazebo. But this could be useful later on.

hsd-dev:
Actually the gazebo tag was already present in the xacro. I just changed its location.

jschleicher:
Do you use prbt_macro.xacro as your main URDF file? You should include the macro from your own xacro instead as shown in [Tutorial 1](https://wiki.ros.org/pilz_robots/Tutorials/ModelYourApplicationWithPRBT). Would that solve your issue?

We want to provide a `prbt_gazebo` package with an example launch file, but unfortunately that is currently not the highest priority.

hsd-dev:
@jschleicher I am using the macro inside my own xacro.
The thing is that the gazebo tag already exists in your current version of macro: https://github.com/PilzDE/pilz_robots/blob/kinetic-devel/prbt_support/urdf/prbt_macro.xacro#L372

The only change in this PR is that I moved the tag from inside the macro to outside the macro, I have not added any new piece of code.

jschleicher:
I cannot reproduce your issue with gazebo7 (Version 7.0.0+dfsg-2).

Could you please try 

```roslaunch prbt_gazebo sim_prbt_gazebo.launch```

on the branch `feature/gazebo-pkg`? This should show the robot in gazebo and I am able to plan and execute motions with the MotionPlanning-rviz-plugin.

agutenkunst:
@ipa-hsd is this PR still relevant?

hsd-dev:
I already spoke to @jschleicher about it. Maybe there some confusion from my side. We can close it.

PR #35, opened 2018-12-11
Refactor/new testutils package

martiniil:
* Makes AsyncTest class available for other packages.
* Refactoring of AsyncTest

### DIFF
--- a/pilz_testutils/CHANGELOG.rst
+++ b/pilz_testutils/CHANGELOG.rst
@@ -1,0 +1,8 @@
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Changelog for package pilz_testutils
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Forthcoming
+-----------
+* Provide class AsyncTest for tests with asynchronous events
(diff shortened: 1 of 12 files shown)


pilz_robots/package.xml
```diff
@@ -18,6 +18,7 @@
   <exec_depend>prbt_support</exec_depend>
   <exec_depend>prbt_ikfast_manipulator_plugin</exec_depend>
   <exec_depend>prbt_moveit_config</exec_depend>
+  <exec_depend>pilz_testutils</exec_depend>
```
jschleicher: Should the meta-package really depend on the test_utils to install the package as dependency?
I'd prefer the package to be `test_depend` in the packages that need it.

jschleicher (approved):
I'd suggest to add a user documentation what this class is useful for (best with some sample code in the class header). Maybe not part of this PR but at least with the next release we should
- [ ] create a wiki page for the new package
- [ ] Add a Readme.md to the test_utils package explaining the purpose for ROS users that want to use the class inside their tests (and link the readme from the wiki page).

Please feel free to merge here and file a follow up PR. Therefore Approving.

pilz_robots/package.xml
martiniil: Done.

PR #40, opened 2018-12-20
Set default pipeline to ompl

martiniil:


jschleicher:
Please add a more descriptive commit message and/or add a changelog entry. Probably: "to run with the specified run_depends we cannot default to command_planner"

agutenkunst (approved):

PR #45, opened 2019-01-08
drop duplicate comment in robot.launch

jschleicher:


martiniil:
Thank you @jschleicher. The build failure is due to:

`[WARN] [1546959318.081684, 0.000000]: Controller Spawner couldn't find the expected controller_manager ROS interface.`

This can happen when the launch process is delayed significantly. I will restart the builds.

martiniil (approved):

PR #60, opened 2019-01-17
drop outdated can configuration

jschleicher:
* vl mode isnt supported by the firmware
* publishing safe-torque-off-register 20A0 was meant for auto-enable
  and was superseded by modubs (package prbt_hardware_support)


prbt_support/config/manipulator_driver.yaml
```diff
@@ -25,12 +25,6 @@ defaults:
   eds_pkg: prbt_support
   eds_file: "config/prbt_0_1.dcf"
 
-  #the following scaling is due to 16bit value range limitation of velocity command in vl mode (2)
-  vel_to_device: "rint(rad2deg(vel)*250)"
-  dcf_overlay: # "ObjectID": "ParameterValue" (both as strings)
-    "604Csub1": "1" # vl dimension factor numerator
-    "604Csub2": "24000" # vl dimension factor denominator
-
```
jschleicher, line 33: Actually vl mode *is* supported; but I've never tested, how it works. So imho we should either test and document or drop it.

agutenkunst (approved):
Thank you!

Issue #70, opened 2019-02-14
Gazebo integrationtest errors


martiniil:
### Commit
93454eb1c7e1fe3e46ec47d8d778d73d68722161

### Steps to reproduce
This happened here: https://travis-ci.org/PilzDE/pilz_robots/jobs/492676506
I do not know, how to reproduce it.

### Log

> [ INFO] [1550063952.451245135]: Loading robot model 'prbt'...
[ WARN] [1550063952.451764145]: Skipping virtual joint 'FixedBase' because its child frame 'prbt_base_link' does not match the URDF frame 'world'
[ INFO] [1550063952.452015314]: No root/virtual joint specified in SRDF. Assuming fixed joint
[ INFO] [1550063953.094379308]: Finished loading Gazebo ROS API Plugin.
[ INFO] [1550063953.107568358]: waitForService: Service [/gazebo/set_physics_properties] has not been advertised, waiting...
[ INFO] [1550063953.189882115]: Publishing maintained planning scene on 'monitored_planning_scene'
[ INFO] [1550063953.231851913]: MoveGroup debug mode is OFF
Starting context monitors...
[ INFO] [1550063953.231918243]: Starting scene monitor
[ INFO] [1550063953.278778984]: Listening to '/planning_scene'
[ INFO] [1550063953.278831298]: Starting world geometry monitor
[ INFO] [1550063953.325316757]: Listening to '/collision_object' using message notifier with target frame '/world '
[ INFO] [1550063953.331388117]: Listening to '/planning_scene_world' for planning scene world geometry
[ INFO] [1550063953.345524752]: Listening to '/attached_collision_object' for attached collision objects
Context monitors started.
[ INFO] [1550063953.391006143]: Initializing OMPL interface using ROS parameters
SpawnModel script started
[ INFO] [1550063953.452645766]: Using planning interface 'OMPL'
[ INFO] [1550063953.471948682]: Param 'default_workspace_bounds' was not set. Using default value: 10
[ INFO] [1550063953.473790777]: Param 'start_state_max_bounds_error' was set to 0.1
[ INFO] [1550063953.475083585]: Param 'start_state_max_dt' was not set. Using default value: 0.5
[ INFO] [1550063953.476805438]: Param 'start_state_max_dt' was not set. Using default value: 0.5
[ INFO] [1550063953.478555325]: Param 'jiggle_fraction' was set to 0.05
[...]

PR #71, opened 2019-02-15
Use blank.world in gazebo integrationtest

agutenkunst:
* Targets #70
* Some errors where observered trying to load models (e.g. sun) when running
  in travis. With this change the integrationtest loads the blank.world which
  has no models.


martiniil:
> Invalid roslaunch XML syntax: mismatched tag: line 35, column 4

martiniil (approved):

PR #84, opened 2019-03-20
pass rviz config file path as launch argument

jschleicher:
According to @rfeistenauer it is preferred to hava a rviz-config in
an application package and not just a user-default.
(See https://github.com/ros-planning/moveit/pull/1397)

Re-generated demo.launch from upstream template.


ct2034 (changes requested):
The naming of params handed through launch files should be consistent `(i.e. rviz_config != rviz_config_file)`

ct2034 (approved):
Thanks! That is what I needed :smile:

PR #87, opened 2019-03-26
Add missing dependency on joint_state_controller

martiniil:
Also remove some redundant dependencies.

### DIFF
--- a/prbt_moveit_config/package.xml
+++ b/prbt_moveit_config/package.xml
@@ -24,7 +24,6 @@
   <exec_depend>moveit_ros_visualization</exec_depend>
   <exec_depend>moveit_fake_controller_manager</exec_depend>
   <exec_depend>moveit_simple_controller_manager</exec_depend>
-  <exec_depend>joint_trajectory_controller</exec_depend>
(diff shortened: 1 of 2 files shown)


jschleicher:
Why is `joint_trajectory_controller` redundant? prbt_support starts it in its controller config?! Do you want to move `joint_trajectory_controller` as `<run_depend>` into `prbt_hardware_support`?

jschleicher (changes requested):
Please keep `joint_trajectory_controller` as dependency of prbt_support.

martiniil:
@jschleicher thanks for your comments.
prbt_support starts the `pilz_joint_trajectory_controller` from the dependency `pilz_control` (which of course depends on `joint_trajectory_controller`). That's why I think it was redundant. What do you think?

jschleicher:
Alright, I had a look into prbt_hardware support, but you're right. The controller part is inside pilz_control.

jschleicher (approved):

jschleicher:
Maybe add a comment near the exec_depend of pilz_control.

PR #94, opened 2019-03-29
Fix test_context.launch in prbt_moveit_config

martiniil:
* load joint limits of gripper joints
* hotfix: default joint limits namespace for tests with pilz command_planner

I would aim for a release asap such that [pilz_industrial_motion#131](https://github.com/PilzDE/pilz_industrial_motion/pull/131) is not delayed.

### DIFF
--- a/prbt_moveit_config/launch/test_context.launch
+++ b/prbt_moveit_config/launch/test_context.launch
@@ -19,11 +19,19 @@
 
   <!-- Load updated joint limits (override information from URDF) -->
   <group ns="$(arg robot_description)_planning">
-    <rosparam command="load" file="$(find prbt_moveit_config)/config/joint_limits.yaml"/>
+    <rosparam if="$(eval not gripper)" command="load" file="$(find prbt_moveit_config)/config/joint_limits.yaml"/>
+    <rosparam unless="$(eval not gripper)" command="load"
+              file="$(eval find('prbt_'+ gripper + '_support') + '/config/joint_limits.yaml')" />
+
+    <!-- Load cartesian limits -->
+    <rosparam command="load" file="$(find prbt_moveit_config)/config/cartesian_limits.yaml"/>
   </group>
 
-  <!-- Load updated joint limits (override information from URDF) -->
-  <group ns="$(arg robot_description)_planning">
+  <!-- Load limits again in default namespace (which may not be configurable for planners) -->
+  <group ns="robot_description_planning">
+    <rosparam if="$(eval not gripper)" command="load" file="$(find prbt_moveit_config)/config/joint_limits.yaml"/>
+    <rosparam unless="$(eval not gripper)" command="load"
+              file="$(eval find('prbt_'+ gripper + '_support') + '/config/joint_limits.yaml')" />
     <rosparam command="load" file="$(find prbt_moveit_config)/config/cartesian_limits.yaml"/>
   </group>
 


prbt_moveit_config/launch/test_context.launch
```diff
@@ -19,14 +19,24 @@
...
+              file="$(eval find('prbt_'+ gripper + '_support') + '/config/joint_limits.yaml')" />
   </group>
 
-  <!-- Load updated joint limits (override information from URDF) -->
+  <!-- Load cartesian limits -->
   <group ns="$(arg robot_description)_planning">
     <rosparam command="load" file="$(find prbt_moveit_config)/config/cartesian_limits.yaml"/>
   </group>
 
+  <!-- Load limits again in default namespace (which may not be configurable for planners) -->
```
jschleicher, line 30: Why would you want to duplicate the settings?

Doesn't your test case  pilz_trajectory_generation/test/unittest_trajectory_generator_common.test already set both params by including this launch file twice with different arguments?

prbt_moveit_config/launch/test_context.launch
```diff
@@ -19,14 +19,24 @@
...
   <!-- Load updated joint limits (override information from URDF) -->
   <group ns="$(arg robot_description)_planning">
-    <rosparam command="load" file="$(find prbt_moveit_config)/config/joint_limits.yaml"/>
+    <rosparam if="$(eval not gripper)" command="load" file="$(find prbt_moveit_config)/config/joint_limits.yaml"/>
+    <rosparam unless="$(eval not gripper)" command="load"
+              file="$(eval find('prbt_'+ gripper + '_support') + '/config/joint_limits.yaml')" />
   </group>
 
-  <!-- Load updated joint limits (override information from URDF) -->
+  <!-- Load cartesian limits -->
```
jschleicher: you could combine both joint and cartesian limits into the same `<group>`.

prbt_moveit_config/launch/test_context.launch
martiniil: yes, it is included twice and loading the robot models from different parameters/namespaces works fine, but not the initialization of the `pilz_command_planner`, due to:

> static const std::string PARAM_NAMESPACE_LIMTS = "robot_description_planning";

So I want to have all joint_limits (including gripper) in addition to be stored in the namespace `robot_description_planning`. As long as the models loaded from the urdf are discriminated correctly, this works fine.

martiniil:
Actually, we could avoid this issue simply by omitting the parametrized tests with and without gripper.

However, the file planning_context.launch suggests that you can change the parameter name `robot_description`. But this is not the case, see https://github.com/ros-planning/moveit/blob/ff50476c4070eb86d0a70aa39281d5805db13fa5/moveit_ros/move_group/src/move_group.cpp#L47

> static const std::string ROBOT_DESCRIPTION =
>     "robot_description";  // name of the robot description (a param name, so it can be changed externally)

martiniil:
The same holds for the OMPL planner: `"robot_description"` is declared static const.

jschleicher:
Yep, I hate this parametrized test setup having different robot models loaded at the same time. There should only be one `robot_description` at a time.

As a quick solution I'd suggest splitting this launch file out of `prbt_moveit_config` to not pollute the installed package any longer. Only a tiny part is used inside `prbt_support/test` which currently doesn't declare a test-depend on the moveit package so
* load the setup needed directly inside `prbt_support/test/urdf_tests.test`

* Drop test_context.launch from prbt_moveit_config
* Integrate a modified test_context.launch directly in PilzDE/pilz_industrial_motion#131 (making that PR self-contained without having to wait for pilz_robots). That new launch file would live in the test directory and is not installed.

What do you think?

martiniil:
Closing in favor of #95

Issue #93, opened 2019-03-27
CAN Error SAFE_TORQUE_OFF (ER: 129)


JonathanGruner:
### Commit
Ubuntu 18.04 using ROS melodic.

### Steps to reproduce
How can the issue/bug be reproduced?
1. Start the robot with holding the Enabling switch
2. Release the switch, press it again, release it, ..... 
3. The Error will occur and the brakes won't open again after a few (sometimes even after the first) releases of the switch
4. If I don't release the button (or in my case close and lock the door of the cell in automatic mode) the error won't occur as often as with the other tactic.

### Expected behavior
I expected, that the robot will continue working after the enabling switch is pressed again.

### Observed behavior
After Error I have to reboot the program (not the robot) to make it work again

### Logfiles
CAN Log:
[candump-2019-03-27_105509.log](https://github.com/PilzDE/pilz_robots/files/3012949/candump-2019-03-27_105509.log)

Console Log:
[Terminal output.txt](https://github.com/PilzDE/pilz_robots/files/3012956/Terminal.output.txt)

jschleicher:
The emergency `EMCY: 87#001081 29 03000000` is         `ERROR_SAFE_TORQUE_OFF  //0x29`

ct2034:
This was **not** reproducible with firmware
```
"100 Build:11158 Date:2018-06-07 16:49:55%"
```

JonathanGruner:
I have firmware:
``` "101 Build 11345 Date:2018-12-17 07:33:065" ```
Its different!

ct2034:
In the archive there are three can logs for comparison:
[can_dumps_wireshark.zip](https://github.com/PilzDE/pilz_robots/files/3191531/can_dumps_wireshark.zip)
- `fw100_error129_no_failure_gz.pcapng`: old firmware - error does not appear
- `fw101_error129_failes_gz.pcapng`: new firmware - fails
- `fw101_error129_workaround_gz.pcapng`: with workaround

martiniil:
Closed via #175

Issue #81, opened 2019-03-11
Modbus Timeout kills launch process


JonathanGruner:
### Commit
 05d7a01 

### Steps to reproduce
Start launch file on low-end pc with a large robot model. If the startup process takes more than 5 seconds, the modbus node dies.

### Expected behavior
Node waits until ROS is ready. There should not be a timeout at all, but just a warning to notify the user: "Service xxx not yet available, waiting...".

### Observed behavior
Modbus node dies and takes down ROS.

### Logfiles

```
[ INFO] [1552301651.010005889]: waitForService: Service [/prbt/manipulator_joint_trajectory_controller/hold] has not been advertised, waiting...
process[usb_cam-11]: started with pid [4544]
process[aruco_detect-12]: started with pid [4551]
process[ref_fid_publisher-13]: started with pid [4553]
process[fiducial_tf_converter-14]: started with pid [4567]
[ INFO] [1552301651.281068524]: Loading robot model 'prbt'...
[ WARN] [1552301651.281614825]: Skipping virtual joint 'FixedBase' because its child frame 'prbt_base_link' does not match the URDF frame 'base'
[ INFO] [1552301651.281992211]: No root/virtual joint specified in SRDF. Assuming fixed joint
[ INFO] [1552301651.980708962]: rviz version 1.12.16
[ INFO] [1552301651.980808438]: compiled against Qt version 5.5.1
[ INFO] [1552301651.980838067]: compiled against OGRE version 1.9.0 (Ghadamon)
[ INFO] [1552301652.355542478]: Using fixed control period: 0.010000000
[ INFO] [1552301652.587717879]: Initializing XXX
[ INFO] [1552301652.588118552]: Current state: 1 device error: system:0 internal_error: 0 (OK)
[ INFO] [1552301652.588401682]: Current state: 2 device error: system:0 internal_error: 0 (OK)
EMCY: 83#0090212500000044
EMCY: 84#0090212500000044
EMCY: 85#0090212500000044
[ INFO] [1552301652.812759987]: Stereo is NOT SUPPORTED
[ INFO] [1552301652.813402492]: OpenGl version: 3 (GLSL 1.3).
EMCY: 86#0090212500000044
EMCY: 87#0090212500000044
EMCY: 88#0090212500000044
ER: 33
[ INFO] [1552301653.775363515]: Current state: 2 device error: system:125 internal_error: 0 (OK)
[...]

PR #97, opened 2019-04-03
longer timeout for modbus responses

ct2034:
I tested it with different ethernet adapters: This timeout can always be met.

### DIFF
--- a/prbt_hardware_support/CHANGELOG.rst
+++ b/prbt_hardware_support/CHANGELOG.rst
@@ -2,6 +2,10 @@
 Changelog for package prbt_hardware_support
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+Forthcoming
+-----------
+* increased modbus response timeout to 20ms
+
 0.5.3 (2019-04-24)
 ------------------
 * cleanup CMakeLists of prbt_hardware_support
--- a/prbt_hardware_support/include/prbt_hardware_support/pilz_modbus_read_client.h
+++ b/prbt_hardware_support/include/prbt_hardware_support/pilz_modbus_read_client.h
@@ -109,7 +109,7 @@ private:
   //! Index from which the registers have to be read.
   const uint32_t INDEX_OF_FIRST_REGISTER;
 
-  static constexpr unsigned int RESPONSE_TIMEOUT_IN_MS {10};
+  static constexpr unsigned int RESPONSE_TIMEOUT_IN_MS {20};
 
   static constexpr double MODBUS_RATE_HZ {500};
 


jschleicher (commented):
What about the Changelog?

hslusarek:
If we change this Timeout, do we have to change the time used for the break trajectory? See [here](https://github.com/PilzDE/pilz_robots/blob/5913a9009b8aef642b01c9000c80d1227d9a0643/prbt_hardware_support/include/prbt_hardware_support/sto_modbus_adapter.h#L94).

hslusarek:
Why do you merge it into melodic? Our develop branch is still kinetic until it is officially moved to melodic.

jschleicher:
> If we change this Timeout, do we have to change the time used for the break trajectory? See [here](https://github.com/PilzDE/pilz_robots/blob/5913a9009b8aef642b01c9000c80d1227d9a0643/prbt_hardware_support/include/prbt_hardware_support/sto_modbus_adapter.h#L94).

No, the brake time is determined by the maximum deceleration of the axes and must not be reduced. Where is the overall stop time documented that is set inside the PNOZ safety program (that should be updated)?

jschleicher (changes requested):
What about the Changelog?
Why do you merge into melodic? Would you mind adding a note to the README as in PilzDE/pilz_industrial_motion#150 about the default development branch? Most probably you need a merge of the current kinetic-branch into melodic before changing the default branch...

agutenkunst:
I need to know a bit more about the reason behind this PR...

agutenkunst:
Also see https://github.com/PilzDE/pilz_robots/pull/97 related?

ct2034:
> I need to know a bit more about the reason behind this PR...

@agutenkunst In some configurations, the modbus communication was timing out with the old timeout. This led probably to #81. If the problem in #81 still exists, we will look at it again.
Can you please check @JonathanGruner

JonathanGruner:
Hi I can check it next week, until then I assume increasing the timeout to 20 ms is good enough, there were no modbus problems with that value yet..

jschleicher (approved):
@agutenkunst Thanks for asking about the rationale:

#93 showed modbus disconnects with PSS4000 (obviously slower reactions than the PNOZmulti) and ct2034 experienced disconnects with USB-Ethernet-Adapters (PNOZ-multi but slower than internal pci network adapters). Both issues were solved when increasing the timeout to 20ms. As a follow-up the delay in the safety program needs to be increased by 10ms (and documented) as well!

hslusarek:
@ct2034 Can we merge?

jschleicher:
@hslusarek 
:+1:  from my side. Please merge.

PR #101, opened 2019-04-15
Add gripper brackets definition to prbt.xacro

### LINKED ISSUE #6: Allow user to modify Gripper Brackets urdf
If the user includes pg70 urdf, default gripper brackets are loaded.
It would be desired, to allow the user, to include / attach his own gripper brackets.

Therefore I'd suggest, to move the [brackets-include](https://github.com/PilzDE/prbt_grippers/blob/6111f3116129387304c852438bcced615647099c/prbt_pg70_support/urdf/pg70.urdf.xacro#L32) into from the gripper-urdf into prbt_support. A user writing his own application.urdf.xacro could then easily replace it with custom brackets.

_Originally posted by @rfeistenauer in https://github.com/PilzDE/pilz_templates/pull/1#issuecomment-477983571_

martiniil:
Now a user can include custom brackets when he writes an own urdf. This was not possible since the default brackets were included in the gripper (pg70) urdf. Fixes PilzDE/prbt_grippers#6

### DIFF
--- a/prbt_support/CHANGELOG.rst
+++ b/prbt_support/CHANGELOG.rst
@@ -2,6 +2,11 @@
 Changelog for package prbt_support
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+Forthcoming
+-----------
+* Add gripper brackets definition to prbt.xacro
+* Contributors: Pilz GmbH and Co. KG
+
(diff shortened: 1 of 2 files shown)


agutenkunst:
Thank you. We have to remember to release prbt_grippers the same time we do the next pilz_robots release!

PR #103, opened 2019-04-15
gripper brackets configurable

jschleicher:
* no longer require parent_left and parent_right arguments, but set
  default values from the gripper name
* add optional parameters to configure the size of the brackets

### DIFF
--- a/prbt_support/urdf/prbt.xacro
+++ b/prbt_support/urdf/prbt.xacro
@@ -57,13 +57,13 @@ limitations under the License.
   <!-- If gripper given then... -->
   <xacro:if value="${gripper != ''}">
     <!-- name of the gripper -->
-    <xacro:arg name="gripper_name" default="$(arg robot_prefix)gripper"/>
+    <xacro:property name="gripper_name" value="$(arg robot_prefix)gripper"/>
 
     <xacro:include filename="$(find prbt_${gripper}_support)/urdf/${gripper}.urdf.xacro" />
 
     <!-- Add gripper brackets -->
     <xacro:include filename="$(find prbt_support)/urdf/simple_gripper_brackets.urdf.xacro" />
-    <xacro:simple_gripper_brackets gripper_name="$(arg gripper_name)" />
+    <xacro:simple_gripper_brackets />
   </xacro:if>
 
 </robot>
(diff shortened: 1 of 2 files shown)


prbt_support/urdf/simple_gripper_brackets.urdf.xacro
```diff
@@ -17,7 +17,19 @@ limitations under the License.
 -->
 
 <robot xmlns:xacro="http://www.ros.org/wiki/xacro">
-  <xacro:macro name="simple_gripper_brackets" params="gripper_name parent_left parent_right">
+  <xacro:macro name="simple_gripper_brackets"
+               params="gripper_name
```
martiniil: You could pass gripper_name as default, too: `gripper_name:=^`

martiniil (changes requested):
The parent_left and _right parameters in  prbt.xacro can be omitted now.

prbt_support/urdf/simple_gripper_brackets.urdf.xacro
jschleicher: I don't want to add a default `prbt_` prefix into the gripper brackets macro to keep it robot independent. You could say, it is inside the prbt_support-package, but I'd rather leave it as is right now.

martiniil: I agree, but the pipe is not necessary? So you would just have the additional option to use `gripper_name` from the outer scope. Otherwise it has to be specified.

jschleicher: Ok, I see. Implemented, although I'm not a fan of implicit parameters...

prbt_support/urdf/prbt.xacro
```diff
@@ -57,16 +57,13 @@ limitations under the License.
   <!-- If gripper given then... -->
   <xacro:if value="${gripper != ''}">
     <!-- name of the gripper -->
-    <xacro:arg name="gripper_name" default="$(arg robot_prefix)gripper"/>
+    <xacro:property name="gripper_name" value="$(arg robot_prefix)gripper"/>
 
     <xacro:include filename="$(find prbt_${gripper}_support)/urdf/${gripper}.urdf.xacro" />
```
jschleicher, line 62: Imho we could use this (API-)change, to also make pg70.urdf.xacro a module (in a follow-up-PR). Why is this just a single include without instantiation?

martiniil: I am not sure. Do we follow any convention in the package `prbt_pg70_support`?

martiniil (approved):

PR #106, opened 2019-04-16
cleanup CMakeLists of prbt_hardware_support

### LINKED ISSUE #105: `prbt_hardware_support` does not install any lib
I am working on Ubuntu 16.04 with ROS kinetic and ROS2 crystal. While building `ros1_bridge`, I get an error:

```
--- stderr: ros1_bridge                         
CMake Error at cmake/find_ros1_package.cmake:75 (message):
  pkg-config module 'prbt_hardware_support' failed to find library
  'prbt_hardware_support'
Call Stack (most recent call first):
  CMakeLists.txt:50 (find_ros1_package)
```
Similar to the issue [here](https://github.com/ros2/ros1_bridge/issues/88), `prbt_hardware_support` exports the package as a library, but does not install any libs. By commenting out the [line](https://github.com/PilzDE/pilz_robots/blob/kinetic-devel/prbt_hardware_support/CMakeLists.txt#L36), I am able to build `ros1_bridge`

jschleicher:
Fixes #105
* no library is built, so drop the export
* pilz_testutils is only declared as test_depend, so move the
  find_package call into the testing section

### DIFF
--- a/prbt_hardware_support/CMakeLists.txt
+++ b/prbt_hardware_support/CMakeLists.txt
@@ -8,12 +8,11 @@ add_definitions(-Werror)
 add_definitions(-std=c++11)
 
 find_package(catkin REQUIRED COMPONENTS
+  message_filters
+  message_generation
   roscpp
   std_msgs
   std_srvs
-  message_generation
-  message_filters
-  pilz_testutils
 )
 
 # message generation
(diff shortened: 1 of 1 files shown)


jschleicher:
I think, this cleanup doesn't need a changelog entry, do you agree?

agutenkunst:
Fine with me

PR #107, opened 2019-04-17
Make docu compabtible with both kinetic and melodic

martiniil:
According to PilzDE/pilz_industrial_motion#150.
The docu should be generated from melodic-devel in the future.


jschleicher:
Regardless of #150, this is a good idea.

jschleicher (approved):

agutenkunst (approved):

PR #108, opened 2019-04-17
Set interactive marker size in RViz config

### LINKED ISSUE #4: Missing interactive marker in rviz, when using gripper:=pg70
### Commit
commit 5913a9009b8aef642b01c9000c80d1227d9a0643
and 
commit 393358665f2c9adf06d9ca5748864dc77ee2d867

(from source)

### Steps to reproduce
$ roslaunch prbt_moveit_config moveit_planning_execution.launch gripper:=pg70

### Expected behavior
interactive marker should be at tcp position

### Observed behavior
no interactive marker available

### Logfiles
No difference to log of starting it without the gripper. But without it I get the marker

### Tried so far
Restarting roscore does not help.

I checked into other issues stating the same problem.
https://answers.ros.org/question/115641/moveit-interactive-markers-missing/ (2014) 
they state labeling the end effektor or fixing parent group  may help, but that was no problem so far

in https://github.com/ros-planning/moveit/issues/736 (2018)
this was pointed to world missing, which is not the case, but the referenced issue is all about
the change in quaternion normalization and initialization issue from the start of the year

maybe the warning:
`
[ WARN] [1554452220.703757035]: Interactive marker 'EE:goal_prbt_flange' contains unnormalized quaternions. This warning will only be output once but may be true for others; enable DEBUG messages for ros.rviz.quaternions to see more details.
`
has something to do with it, but it also appears when launching without the gripper ...

jschleicher:
Fixes PilzDE/prbt_grippers#4

### DIFF
--- a/prbt_moveit_config/launch/moveit.rviz
+++ b/prbt_moveit_config/launch/moveit.rviz
@@ -284,7 +284,7 @@ Visualization Manager:
         Colliding Link Color: 255; 0; 0
         Goal State Alpha: 0.699999988
         Goal State Color: 250; 128; 0
-        Interactive Marker Size: 0
+        Interactive Marker Size: 0.25
         Joint Violation Color: 255; 0; 255
         Planning Group: manipulator
         Query Goal State: true


agutenkunst:
Thank you!

PR #113, opened 2019-04-24
0.5.3

ct2034:
For some reason, I can not push to melodic-devel. So I have to do this for the release


martiniil (approved):
@ct2034 The IK plugin update should be included in the release, too.

PR #114, opened 2019-04-25
WIP: Generalized modbus client

ct2034:
The idea is to provide a Modbus client that can read multiple different registers from the same master. These registers can have individual topics and rates.

This is WIP. We have to discuss the details. Probably this can also be extended to write. And to read registers other than the holding register.


agutenkunst:
Closing for now.

PR #121, opened 2019-05-07
do not install test headers

jschleicher:
1. The headers are not needed at runtime
2. Pollutes code_coverage report


martiniil (approved):
Approved. But how does this pollute the coverage report?

PR #126, opened 2019-05-14
drop rviz exec_depend from prbt_support

jschleicher:
To avoid GUI-dependencies in non-Gui-packages. `prbt_support` is installable on embedded hardware and should not draw in GUI-programs.

Open for discussion / My proposal:

- [x] The launch file `test_urdf.launch` requires RVIZ. ~~Should it be moved into `prbt_moveit_config`? Or a note in Readme.md that RViZ is recommended but not strictly required and auto-installed?~~ I've moved it into the tests directory so that is no longer installed and excluded from `roslaunch_add_file_check`
- [x] acceptance-tests (gripper and tcp) use RViz via test_urdf.launch. ~~Maybe a reference that RViZ has to be installed should be added there?~~ RViZ is now listed as `<test_depend>`.


jschleicher:
Instead of open questions, added a proposal for easier discussion.

If you like, I could add `test_urdf.launch` to [PilzDE/pilz_templates](https://github.com/PilzDE/pilz_templates/), maybe it is useful for testing own robot descriptions but I do not see any need to install it in `prbt_support`.

jschleicher:
@agutenkunst ping

ct2034 (approved):
good suggestion :+1:

PR #130, opened 2019-05-22
Instantiate pg70 xacro macro

martiniil:
Due to forthcoming changes in prbt_grippers: PilzDE/prbt_grippers#9

### DIFF
--- a/prbt_support/urdf/prbt.xacro
+++ b/prbt_support/urdf/prbt.xacro
@@ -59,7 +59,11 @@ limitations under the License.
     <!-- name of the gripper -->
     <xacro:property name="gripper_name" value="$(arg robot_prefix)gripper"/>
 
+    <!-- load gripper macro -->
     <xacro:include filename="$(find prbt_${gripper}_support)/urdf/${gripper}.urdf.xacro" />
+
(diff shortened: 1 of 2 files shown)


prbt_support/urdf/prbt.xacro
```diff
@@ -59,8 +59,14 @@ limitations under the License.
     <!-- name of the gripper -->
     <xacro:property name="gripper_name" value="$(arg robot_prefix)gripper"/>
 
+    <!-- load gripper macro -->
     <xacro:include filename="$(find prbt_${gripper}_support)/urdf/${gripper}.urdf.xacro" />
 
+    <!-- instantiate gripper -->
+    <xacro:if value="${gripper == 'pg70'}">
+      <xacro:pg70 parent="$(arg robot_prefix)flange" />
```
jschleicher: What about naming the macro `prbt_gripper` instead of hard-coding its name? That approach would be generic: include `${gripper}.urdf.xacro` which defines the macro. Instantiate it with one generic macro name.

ct2034 (commented):
Tested it. Looks good.

ct2034 (approved):
Tested it. Looks good.

PR #122, opened 2019-05-07
Wait for required services instead of aborting

jschleicher:
See example for *_DELAYED_THROTTLE in
http://wiki.ros.org/roscpp/Overview/Logging

Fixes #81

Please especially review the test cases I've changed: Instead of `EXPECT_THROW` they now wait for the service in seperate thread.

PR #135, opened 2019-05-29
Add firmeware version logging at start-up

hslusarek:
To have a more complete picture of the system, in case a bug occurs,
the firmeware version of each joint is displayed at start-up on the
console.

### DIFF
--- a/prbt_hardware_support/include/prbt_hardware_support/system_info.h
+++ b/prbt_hardware_support/include/prbt_hardware_support/system_info.h
@@ -1,0 +1,68 @@
+/*
+ * Copyright (c) 2019 Pilz GmbH & Co. KG
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef PRBT_HARDWARE_SUPPORT_SYSTEM_INFO_H
+#define PRBT_HARDWARE_SUPPORT_SYSTEM_INFO_H
+
+#include <ros/ros.h>
+
+#include <string>
+#include <vector>
+#include <map>
+
+#include <canopen_chain_node/GetObject.h>
+
+namespace prbt_hardware_support
+{
+
+//! key = joint_name
+//! value = firmware version of joint
+using FirmwareCont = std::map<std::string, std::string>;
+
+/**
+ * @brief Provides easy access to system information which are of importance
+ * when analyzing bugs.
+ */
+class SystemInfo
+{
+public:
+  SystemInfo(ros::NodeHandle& nh);
+
+  /**
+   * @returns a container comprised of the firmware version of each joint
+   * of the robot.
+   */
+  FirmwareCont getFirmwareVersions();
+
+  /**
+   * @returns the firmware version of the specified joint.
+   */
+  std::string getFirmwareVersionOfJoint(const std::string& joint_name);
+
+private:
+  /**
+   * @returns the names of all joints/nodes.
+   */
+  static std::vector<std::string> getNodeNames(const ros::NodeHandle &nh);
+
+private:
+  const std::vector<std::string> joint_names_;
+  ros::ServiceClient canopen_srv_get_client_;
+};
+
+} // namespace prbt_hardware_support
+#endif // PRBT_HARDWARE_SUPPORT_SYSTEM_INFO_H
--- a/prbt_hardware_support/src/system_info.cpp
+++ b/prbt_hardware_support/src/system_info.cpp
@@ -1,0 +1,96 @@
+/*
+ * Copyright (c) 2019 Pilz GmbH & Co. KG
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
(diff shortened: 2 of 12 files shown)


jschleicher:
Great to have version information!

Mayby it is worth discussing, if this would be a nice upstream feature (https://github.com/ros-industrial/ros_canopen)? Instead of calling as a service, ros_canopen could read this during initialization itself?

prbt_hardware_support/include/prbt_hardware_support/system_info.h
```diff
@@ -0,0 +1,68 @@
...
+
+namespace prbt_hardware_support
+{
+
+//! key = joint_name
+//! value = firmeware version of joint
+using FirmewareCont = std::map<std::string, std::string>;
+
+/**
+ * @brief Provides easy access to system information which are of important
```
martiniil: important -> importance

prbt_hardware_support/include/prbt_hardware_support/system_info.h
```diff
@@ -0,0 +1,68 @@
...
+  FirmewareCont getFirmewareVersions();
+
+  /**
+   * @returns the firmeware version of the specified joint.
+   */
+  std::string getFirmewareVersionOfJoint(const std::string& joint_name);
+
+private:
+  /**
+   * @returns the names of all joints/nodex.
```
martiniil: typo: nodex -> nodes

prbt_hardware_support/src/system_info.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include <prbt_hardware_support/system_info_exception.h>
+
+namespace prbt_hardware_support
+{
+
+static const std::string CANOPEN_GETOBJECT_SERVICE_NAME{"/prbt/driver/get_object"};
+static const std::string CANOPEN_NODES_PARAMETER_NAME{"/prbt/driver/nodes"};
+static const std::string JOINT_STATE_TOPIC {"/joint_states"};
+static constexpr double WAIT_FOR_SERVICE_TIMEOUT_S{3.0};
+static constexpr double JOINT_STATES_TOPIC_TIMEOUT{3.0};
```
martiniil: Why not wait longer? I see no reason to make it that restrictive.

prbt_hardware_support/src/system_info.cpp
```diff
@@ -0,0 +1,104 @@
...
+SystemInfo::SystemInfo(ros::NodeHandle &nh)
+    : joint_names_( getNodeNames(nh) )
+{
+  // Wait till CAN is up and running.
+  // Reason: If the first CAN service call happens before
+  // the CAN is fully initialized, the CAN will not start properly.
+  auto joint_state_msg {ros::topic::waitForMessage<sensor_msgs::JointState>(JOINT_STATE_TOPIC, ros::Duration(JOINT_STATES_TOPIC_TIMEOUT))};
+  if (joint_state_msg == nullptr)
+  {
+    throw SystemInfoException("CAN is not up and running.");
```
martiniil: I would write: `"Could not read from /joint_states topic. Probably CAN is not up and running."`

prbt_support/launch/robot.launch
```diff
@@ -63,6 +63,8 @@
       <arg name="safety_hw" value="$(arg sto)" />
       <arg name="api_spec_file" value="$(find prbt_hardware_support)/config/modbus_api_spec_$(arg sto).yaml" />
     </include>
+    <node ns="prbt" pkg="prbt_hardware_support" name="system_info_node" type="system_info_node"
+          output="screen" />
```
martiniil: Is there a specific reason you put the node it inside of   `<group if="$(eval sto == 'pnoz')" >`?
If not I would prefer it on the outside.

prbt_hardware_support/test/unittests/unittest_system_info.cpp
```diff
@@ -0,0 +1,243 @@
...
+TEST_F(SystemInfoTests, testServiceResponseFalse)
+{
+  EXPECT_CALL(*this, executeGetObject(_,_))
+      .Times(1)
+      .WillRepeatedly(testing::Invoke(
+                        [](GetObject::Request&, GetObject::Response& res){
+                        res.success = false;
+                        return true;
+                      }
+                      ));
```
martiniil, line 229: Looks nice and clean! By the way, this is equivalent:
```
GetObject::Response response;
response.success = false;
EXPECT_CALL(*this, executeGetObject(_,_))
  .WillOnce(DoAll(SetArgReferee<1>(response), Return(true)));
```

martiniil (changes requested):
Please also fix the spelling of firmware (_firmeware_ is wrong) in all files.

prbt_hardware_support/src/system_info.cpp
hslusarek: Yes. I would agree. However, then the tests will take forever ;)

jschleicher: As discussed in #81 it is always a bad idea, to abort node startup with such a fixed timeout. Please refactor as in #122.

hslusarek: > As discussed in #81 it is always a bad idea, to abort node startup with such a fixed timeout. Please refactor as in #122.

I do not want to disagree but in my opinion #122 is even worse especially the lines [here](https://github.com/PilzDE/pilz_robots/blob/de39cc0f9bedeefa85efe822ef25daece1b59e5e/prbt_hardware_support/src/sto_modbus_adapter.cpp#L45) because that way you might end up with a system which will never reach a proper state instead of failing. Or did I miss something?

jschleicher: Yep. As discussed there: RQT and other ros-tools use such a blocking approach. They print a message "waiting for topic ..." and block. That's common in "the ROS world" and we should imho use that design as well.
It has the advantage that the user can split launch files and start different nodes at different times (other nodes waiting). An exception doesn't allow such a use (even worse in the case of the modbus node, that is marked as `required` and thus kills all other nodes as well).
Of course you have to check for `ros::ok()` to allow aborting the loop. Please see also the example in [ros wiki](http://wiki.ros.org/roscpp/Overview/Logging).

hslusarek: > As discussed in #81 it is always a bad idea, to abort node startup with such a fixed timeout. Please refactor as in #122.

Done.

PR #142, opened 2019-06-13
Feature/operation mode

agutenkunst:


prbt_hardware_support/include/prbt_hardware_support/modbus_api_definitions.h
```diff
@@ -25,15 +25,10 @@ namespace prbt_hardware_support
...
-  // Must hold for every version
-  static constexpr uint32_t MODBUS_REGISTER_API {513};
-
   namespace v2
   {
-    static constexpr uint32_t MODBUS_REGISTER_STO {512};
-
-    static constexpr uint16_t MODBUS_STO_CLEAR_VALUE  {1};
-    static constexpr uint16_t MODBUS_STO_ACTIVE_VALUE {0};
+    static constexpr uint16_t MODBUS_STO_CLEAR_VALUE  {1}; // TODO @agu move into ApiSpec
```
hslusarek: When is this happening? And are you responsible for it or everyone?

prbt_hardware_support/README.md
```diff
@@ -64,3 +64,7 @@ The ``ModbusAdapterBrakeTestNode`` is noticed via the topic `/pilz_modbus_node/m
 ## BraketestExecutorNode
 The ``BraketestExecutorNode`` offers the `/execute_braketest` service which, in interaction with the ``CanOpenBraketestAdapter``,
 executes a braketest on each drive of the manipulator.
+
+## ModbusAdapterOperationModeNode
+The ``ModbusAdapterOperationModeNode`` offers the `/get_operation_mode` service for accessing the currently
```
hslusarek: Should we describe which operation mode exist in general and what these modi entail?

prbt_hardware_support/include/prbt_hardware_support/adapter_operation_mode.h
```diff
@@ -0,0 +1,58 @@
...
+{
+
+/**
+ * @brief Offers a service with information on the active operation mode
+ */
+class AdapterOperationMode
+{
+public:
+  AdapterOperationMode(ros::NodeHandle& nh);
+
```
hslusarek, line 37: You have a virtual function but missing a virtual destructor. Please add the virtual destructor. You can write something like ```virtual ~AdapterOperationMode() = default;```
A side note: You can avoid these kind of errors with a properly configured IDE ;)

prbt_hardware_support/include/prbt_hardware_support/adapter_operation_mode.h
```diff
@@ -0,0 +1,58 @@
...
+protected:
+	void init();
+  virtual void updateOperationMode(OperationMode operation_mode);
+  bool getOperationMode(GetOperationMode::Request&, GetOperationMode::Response& response);
+
+private:
+	//! Is the node initialized?
+	bool initialized_;
```
hslusarek: A small side note: Your code formatting is totally off.

prbt_hardware_support/src/adapter_operation_mode.cpp
```diff
@@ -0,0 +1,56 @@
...
+	if(!initialized_) {
+		init();
+		initialized_ = true;
+	}
+}
+
+bool AdapterOperationMode::getOperationMode(GetOperationMode::Request& req,
+				                            GetOperationMode::Response& res)
+{
+	res.mode.value = static_cast<int>(current_operation_mode_);
```
hslusarek: Incorrect cast. Cast to `int8_t`.

prbt_hardware_support/src/adapter_operation_mode.cpp
```diff
@@ -0,0 +1,56 @@
...
+	current_operation_mode_ = mode;
+
+	// when the first data is received, the node is initialized (i.e. the service advertised)
+	if(!initialized_) {
+		init();
+		initialized_ = true;
+	}
+}
+
+bool AdapterOperationMode::getOperationMode(GetOperationMode::Request& req,
```
hslusarek: `req` is unused. Please comment with `/* req */`

prbt_hardware_support/include/prbt_hardware_support/opertion_modes.h
```diff
@@ -0,0 +1,57 @@
...
+
+#ifndef OPERATION_MODE_H
+#define OPERATION_MODE_H
+
+#include <iostream>
+
+namespace prbt_hardware_support
+{
+
+enum class OperationMode
```
hslusarek: This is actually a code duplication from the code which is generated from `OperationModes.msg`. Because the generated code contains an `enum` with the values: `T1=1`, `T2=2`, `AUTO=3`, `UNKNOWN=0`. I understand that you probably want to use a proper type safe data type. In this case, at least add an unittest that checks that the constants in `OpertationModes.msg` and in `OperationMode` are in sync by simply checking that the int values of T1, T2, Auto and unknown are the same.

prbt_hardware_support/include/prbt_hardware_support/opertion_modes.h
```diff
@@ -0,0 +1,57 @@
+  /*
```
hslusarek: The file name is missing an "a".

prbt_hardware_support/include/prbt_hardware_support/opertion_modes.h
hslusarek: Documentation for modi missing.

prbt_hardware_support/src/adapter_operation_mode.cpp
```diff
@@ -0,0 +1,56 @@
...
+	                                              &AdapterOperationMode::getOperationMode,
+	                                                      this);
+}
+
+void AdapterOperationMode::updateOperationMode(OperationMode mode)
+{
+	ROS_INFO_STREAM("Mode switched: " << current_operation_mode_ << " -> " << mode);
+	current_operation_mode_ = mode;
+
+	// when the first data is received, the node is initialized (i.e. the service advertised)
```
hslusarek: Say what it is, namely a "lazy initialization".

prbt_hardware_support/src/adapter_operation_mode.cpp
```diff
@@ -0,0 +1,56 @@
...
+	if(!initialized_) {
```
hslusarek: You could make the `operation_mode_server` variable to a `boost::optional` variable. This would remove the need for the error-prone `initialized_` variable.

prbt_hardware_support/include/prbt_hardware_support/adapter_operation_mode.h
```diff
@@ -0,0 +1,58 @@
...
+ * @brief Offers a service with information on the active operation mode
+ */
+class AdapterOperationMode
+{
+public:
+  AdapterOperationMode(ros::NodeHandle& nh);
+
+protected:
+	void init();
+  virtual void updateOperationMode(OperationMode operation_mode);
```
hslusarek: Nobody actually overrides this function. So does it really need to be `virtual`?

prbt_hardware_support/include/prbt_hardware_support/modbus_adapter_operation_mode.h
```diff
@@ -0,0 +1,69 @@
...
+#include <prbt_hardware_support/modbus_msg_operation_mode_wrapper.h>
+
+namespace prbt_hardware_support
+{
+
+/**
+ * @brief Listens to the modbus_read topic and offers a service
+ * informing about the active operation mode
+ */
+class ModbusAdapterOperationMode : public AdapterOperationMode
```
hslusarek, line 41: I actually see no reason why `ModbusAdapterOperationMode` needs to be derived from `AdapterOperationMode`. Only the naming, which in my opinion is not correct, might suggest something like this. `ModbusAdapterOperationMode` is the actual adapter which takes a Modbus-Msg and converts it with the help of the `ModbusMsgOperationModeWrapper` in an OperationMode data type. The `AdapterOperationMode` class is actually not an adapter, but a "publisher" of sorts.

I suggest to make the `AdapterOperationMode` class not a base class of `ModbusAdapterOperationMode` but a member. This should also simply testing because then the `AdapterOperationMode` can be injected into the `ModbusAdapterOperationMode` class.

prbt_hardware_support/include/prbt_hardware_support/adapter_operation_mode.h
```diff
@@ -0,0 +1,58 @@
...
+#include <prbt_hardware_support/GetOperationMode.h>
+#include <prbt_hardware_support/opertion_modes.h>
+
+namespace prbt_hardware_support
+{
+
+/**
+ * @brief Offers a service with information on the active operation mode
+ */
+class AdapterOperationMode
```
hslusarek, line 32: Please rename this class. It is not a adapter but simply a publisher. See also my comments in the `ModbusAdapterOperationMode` class.

prbt_hardware_support/include/prbt_hardware_support/modbus_adapter_operation_mode.h
```diff
@@ -0,0 +1,69 @@
...
+private:
+  void modbusInMsgCallback(const ModbusMsgInStampedConstPtr& msg);
+  void internalMsgCallback(const ModbusMsgOperationModeWrapper& msg);
+
+private:
+  //! Filters consecutive messages with the same timestamp.
+  //! Passed messages are redirected to the OperationModeFilter.
+  std::shared_ptr< message_filters::UpdateFilter<ModbusMsgInStamped> > update_filter_;
+
+  //! Filters messages with a change in brake test state.
```
hslusarek: Update docu.

prbt_hardware_support/include/prbt_hardware_support/modbus_adapter_operation_mode.h
```diff
@@ -0,0 +1,69 @@
...
+
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MODBUS_ADAPTER_OPERATION_MODE_H
+#define MODBUS_ADAPTER_OPERATION_MODE_H
+
+#include <memory>
+#include <string>
```
hslusarek: Remove.

prbt_hardware_support/include/prbt_hardware_support/modbus_adapter_operation_mode.h
```diff
@@ -0,0 +1,69 @@
...
+#include <ros/ros.h>
+#include <message_filters/subscriber.h>
+
+#include <prbt_hardware_support/adapter_operation_mode.h>
+
+#include <prbt_hardware_support/operation_mode_filter.h>
+#include <prbt_hardware_support/ModbusMsgInStamped.h>
+#include <prbt_hardware_support/update_filter.h>
+#include <prbt_hardware_support/modbus_msg_operation_mode_wrapper.h>
+
```
hslusarek, line 33: Add `#include <prbt_hardware_support/modbus_api_spec.h>`

prbt_hardware_support/include/prbt_hardware_support/modbus_api_spec.h
```diff
@@ -18,27 +18,110 @@
...
+      throw ModbusApiSpecException(e.what());
+    }
+  }
+
+private:
+  std::map<std::string, unsigned int> register_mapping_;
 };
 
+//! Simple typedef for class like usage
+typedef ModbusApiSpecTemplated<ros::NodeHandle> ModbusApiSpec;
```
hslusarek: Write `typedef ModbusApiSpecTemplated<> ModbusApiSpec;` then your default type will be used and you do not have to duplicate the default type.

prbt_hardware_support/include/prbt_hardware_support/modbus_api_spec.h
```diff
@@ -18,27 +18,110 @@
...
+    }
+
+    for (auto rpci = rpc.begin(); rpci != rpc.end(); ++rpci)
+    {
+      int value = rpci->second;
+      setRegisterDefinition(rpci->first.c_str(), static_cast<unsigned int>(value));
+    }
+  }
+
+  bool hasRegisterDefinition(const std::string &key) const
```
hslusarek: `inline` function.

prbt_hardware_support/include/prbt_hardware_support/modbus_api_spec.h
```diff
@@ -18,27 +18,110 @@
...
+  {
+    return register_mapping_.find(key) != register_mapping_.end();
+  }
 
-  const unsigned int version_register_;
-  const unsigned int braketest_register_;
+  void setRegisterDefinition(const std::string &key, unsigned int value)
```
hslusarek: `inline` function.

prbt_hardware_support/include/prbt_hardware_support/modbus_msg_wrapper.h
```diff
@@ -0,0 +1,128 @@
...
+class ModbusMsgWrapper
+{
+public:
+  /**
+   * @brief Construct a new Modbus Msg Wrapper object
+   *
+   * @throw ModbusMsgWrapperException if no version can be found in the message and thus interpretation is impossible
+   */
+  ModbusMsgWrapper(const ModbusMsgInStampedConstPtr& modbus_msg_raw, const ModbusApiSpec& api_spec);
+
```
hslusarek, line 42: `virtual` destructor missing although class has virtual functions.

prbt_hardware_support/include/prbt_hardware_support/modbus_msg_wrapper.h
```diff
@@ -0,0 +1,128 @@
...
+  virtual bool isDisconnect() const;
+
+protected:
+
+  /**
+   * @brief Check if a certain holding register is define in the Modbus message.
+   *
+   * @returns true if the message has the register defined, otherwise false.
+   */
+  virtual bool hasRegister(const ModbusMsgInStampedConstPtr& modbus_msg_raw,
```
hslusarek: Why do you have member `msg_` of type `ModbusMsgInStampedConstPtr` (initialized in the construtor), if you pass the message `modbus_msg_raw` in as a argument? Please remove the argument `modbus_msg_raw` from the function signature and use the member `msg_` instead. These statement holds for **ALL** functions in this class. So, please also check the other functions.

Also: Remove the `virtual` from **ALL** functions which are not overwritten by the derived classes which is basically true for all functions in this class.

prbt_hardware_support/include/prbt_hardware_support/modbus_msg_wrapper.h
```diff
@@ -0,0 +1,128 @@
...
+  ModbusMsgInStampedConstPtr msg_;
+  const ModbusApiSpec api_spec_;
+};
+
+inline ModbusMsgWrapper::ModbusMsgWrapper(const ModbusMsgInStampedConstPtr& modbus_msg_raw,
+                                          const ModbusApiSpec& api_spec):
+  msg_(modbus_msg_raw),
+  api_spec_(api_spec)
+{
+  if (isDisconnect()) // TODO  @agu find a better way
```
hslusarek: When will we do this?

prbt_hardware_support/include/prbt_hardware_support/modbus_msg_wrapper.h
```diff
@@ -0,0 +1,128 @@
...
+  virtual uint16_t getRegister(const ModbusMsgInStampedConstPtr& modbus_msg_raw,
+                               uint32_t reg) const;
+
+  /**
+   * @brief Check if the modbus_msg contains the API version.
+   */
+  virtual bool hasVersion(const ModbusMsgInStampedConstPtr& modbus_msg_raw) const;
+
+protected:
+  ModbusMsgInStampedConstPtr msg_;
```
hslusarek: Add the modifier `const` to the member.
Also: The member can be private, because the derived classes have access to everything they need via `has/getRegister()` function.

prbt_hardware_support/include/prbt_hardware_support/modbus_msg_wrapper.h
```diff
@@ -0,0 +1,128 @@
...
+#include <prbt_hardware_support/modbus_msg_wrapper_exception.h>
+
+namespace prbt_hardware_support
+{
+
+/**
+ * @brief Wrapper class to add semantic to a raw ModbusMsgInStamped
+ *
+ * Allows to easy access to the content behind a raw modbus message
+ * which is assumed to contain data about the operation mode.
```
hslusarek: Update docu.

prbt_hardware_support/include/prbt_hardware_support/modbus_msg_operation_mode_wrapper.h
```diff
@@ -0,0 +1,95 @@
...
+}
+
+inline bool ModbusMsgOperationModeWrapper::hasOperationMode(const ModbusMsgInStampedConstPtr& modbus_msg_raw) const
+{
+  return hasRegister(modbus_msg_raw, api_spec_.getRegisterDefinition(modbus_api_spec::OPERATION_MODE));
+}
+
+inline OperationMode ModbusMsgOperationModeWrapper::getOperationMode() const
+{
+    switch(getRegister(msg_, api_spec_.getRegisterDefinition(modbus_api_spec::OPERATION_MODE))){
```
hslusarek: Like before, this function must be synced via tests with the operation mode values in `OperationMode.msg` and the values in the `OperationMode` enum class.

prbt_hardware_support/include/prbt_hardware_support/modbus_msg_operation_mode_wrapper.h
```diff
@@ -0,0 +1,95 @@
...
+ * Allows to easy access to the content behind a raw modbus message
+ * which is assumed to contain data about the operation mode.
+ */
+class ModbusMsgOperationModeWrapper : public ModbusMsgWrapper
+{
+public:
+  ModbusMsgOperationModeWrapper(const ModbusMsgInStampedConstPtr& modbus_msg_raw, const ModbusApiSpec& api_spec);
+
+  /**
+   * @brief Get the brake test required flag from the Modbus message.
```
hslusarek: Update docu.

prbt_hardware_support/include/prbt_hardware_support/modbus_msg_operation_mode_wrapper.h
```diff
@@ -0,0 +1,95 @@
...
+   *
+   * @return true if the a brake test is required, otherwise false.
+   */
+  OperationMode getOperationMode() const;
+
+private:
+
+  /**
+   * @brief Check if the message contains a brake test required definition.
```
hslusarek: Update docu.

prbt_hardware_support/include/prbt_hardware_support/modbus_msg_in_utils.h
```diff
@@ -51,6 +53,96 @@ static ModbusMsgInStamped* createDefaultModbusMsgIn(const uint32_t& offset,
...
+ModbusMsgInStampedPtr createDefaultOpModeModbusMsg(unsigned int operation_mode,
+                                                    unsigned int modbus_api_version,
+                                                    uint32_t operation_mode_index,
+                                                    uint32_t version_index)
+{
+  uint32_t first_index_to_read{std::min(operation_mode_index, version_index)};
+  uint32_t last_index_to_read{std::max(operation_mode_index, version_index)};
+  static int msg_time_counter{1};
+  std::vector<uint16_t> tab_reg(last_index_to_read - first_index_to_read + 1);
+  tab_reg[version_index - first_index_to_read] = modbus_api_version;
```
hslusarek: Warning: "implicit conversion loses integer precision"

prbt_hardware_support/include/prbt_hardware_support/modbus_msg_in_utils.h
```diff
@@ -51,6 +53,96 @@ static ModbusMsgInStamped* createDefaultModbusMsgIn(const uint32_t& offset,
...
+  tab_reg[operation_mode_index - first_index_to_read] = operation_mode;
```
hslusarek: Warning: "implicit conversion loses integer precision"

prbt_hardware_support/include/prbt_hardware_support/modbus_msg_in_utils.h
```diff
@@ -51,6 +53,96 @@ static ModbusMsgInStamped* createDefaultModbusMsgIn(const uint32_t& offset,
...
+      }
+    }
+
+    ModbusMsgInBuilder& setSto(uint16_t sto){
+      setRegister(api_spec_.getRegisterDefinition(modbus_api_spec::STO), sto);
+      return *this;
+    }
+
+    ModbusMsgInBuilder& setOperationMode(OperationMode mode){
+      setRegister(api_spec_.getRegisterDefinition(modbus_api_spec::OPERATION_MODE), static_cast<int>(mode));
```
hslusarek: Incorrect cast. Cast to `uint16_t`.

prbt_hardware_support/include/prbt_hardware_support/modbus_msg_in_utils.h
```diff
@@ -51,6 +53,96 @@ static ModbusMsgInStamped* createDefaultModbusMsgIn(const uint32_t& offset,
...
+      for(auto reg : register_values_)
+      {
+        tab_reg[reg.first - first_index_to_read] = reg.second;
+      }
+
+      ModbusMsgInStampedPtr msg { new ModbusMsgInStamped() };
+
+      msg->holding_registers.layout.data_offset = first_index_to_read;
+      msg->holding_registers.layout.dim.push_back(std_msgs::MultiArrayDimension());
+      msg->holding_registers.layout.dim[0].size = tab_reg.size();
```
hslusarek, line 126: Warning: "implicit conversion loses integer precision"

prbt_hardware_support/include/prbt_hardware_support/modbus_msg_in_utils.h
```diff
@@ -51,6 +53,96 @@ static ModbusMsgInStamped* createDefaultModbusMsgIn(const uint32_t& offset,
...
+      msg->holding_registers.layout.dim[0].label = "Data in holding register";
+
+      msg->holding_registers.data = tab_reg;
+      msg->header.stamp = ros::Time::now();
+      msg->disconnect.data = false;
+
+      return msg;
+    }
+
+    void setRegister(unsigned int register_n, uint16_t value){
```
hslusarek: `inline` function

prbt_hardware_support/include/prbt_hardware_support/modbus_msg_sto_wrapper.h
```diff
@@ -31,17 +32,10 @@ namespace prbt_hardware_support
  * Allows to easy access to the content behind a raw modbus message which is assumed to contain
  * data about STO clearance.
  */
-class ModbusMsgStoWrapper
+class ModbusMsgStoWrapper : public ModbusMsgWrapper // TODO @agu place this under ModbusMsgWrapper
```
hslusarek: What about this TODO?

prbt_hardware_support/include/prbt_hardware_support/modbus_msg_sto_wrapper_exception.h
```diff
@@ -4,18 +4,20 @@
 #include <string>
```
hslusarek, line 21: License missing.

prbt_hardware_support/include/prbt_hardware_support/operation_mode_filter.h
```diff
@@ -0,0 +1,115 @@
...
+private:
+  void cb(const EventType &evt)
+  {
+    try
+    {
+      prbt_hardware_support::ModbusMsgOperationModeWrapper msg(evt.getMessage(), api_spec_);
+
+      if (!msg.isDisconnect())
+      {
+        if (msg.getOperationMode() == last_operation_mode_ && !first_call_)
```
hslusarek: Avoid `if` nesting

prbt_hardware_support/include/prbt_hardware_support/operation_mode_filter.h
```diff
@@ -0,0 +1,115 @@
...
+      ROS_ERROR_STREAM(ex.what() << "\nCan not interpret Modbus message as OperationMode.");
+      return;
+    }
+
+    this->signalMessage(evt);
+  }
+
+  Connection incoming_connection_;
+
+  prbt_hardware_support::OperationMode last_operation_mode_{prbt_hardware_support::OperationMode::UNKNOWN};
```
hslusarek: You could make this member to a `boost::optional` variable. This would remove the need for the error-prone `first_call_` variable.

prbt_hardware_support/include/prbt_hardware_support/modbus_msg_operation_mode_wrapper_exception.h
```diff
@@ -0,0 +1,42 @@
...
+#include <stdexcept>
+
+namespace prbt_hardware_support
+{
+  /**
+   * @brief Expection thrown upon construction of ModbusMsgOperationModeWrapper
+   * of the message does not contain the required information.
+   *
+   */
+  class ModbusMsgOperationModeWrapperException : public std::runtime_error
```
hslusarek: Why is this exception class not derived from `ModbusMsgWrapperException`?

prbt_hardware_support/msg/OperationModes.msg
```diff
@@ -0,0 +1,24 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+int8 value
+
+# values of CANopen object brake_test_status, see
```
hslusarek: Update docu

prbt_hardware_support/msg/OperationModes.msg
```diff
@@ -0,0 +1,24 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+int8 value
```
hslusarek, line 17: Add docu

prbt_hardware_support/src/modbus_adapter_operation_mode.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+#include <prbt_hardware_support/modbus_topic_definitions.h>
+
+
+namespace prbt_hardware_support
+{
+
+static constexpr int DEFAULT_QUEUE_SIZE_MODBUS {1};
+
+static constexpr unsigned int MODBUS_API_VERSION_REQUIRED {2};
```
hslusarek: Remove. Not used.

prbt_hardware_support/test/integrationtests/integrationtest_operation_mode.cpp
```diff
@@ -0,0 +1,216 @@
...
+    {
+      return ::testing::AssertionFailure() << "Could not call service";
+    }
+    if(srv.response.mode.value == expectation){
+      return ::testing::AssertionSuccess() << "It took " << i+1 << " tries for the service call.";
+    }
+    sleep(1); // This then may take {retries*1}seconds.
+  }
+  return ::testing::AssertionFailure() << "Did not get expected operation mode:"
+                                       << "Actual: " << (int)srv.response.mode.value
```
hslusarek: Old style casts use `static_cast<...>` instead.

prbt_hardware_support/test/integrationtests/integrationtest_operation_mode.cpp
```diff
@@ -0,0 +1,216 @@
...
+ * @note the test is derived from testing::AsyncTest which allows the asynchronous processes to re-sync
+ */
+class OperationModeIntegrationTest : public testing::Test, public testing::AsyncTest
+{
+protected:
+  ros::NodeHandle nh_;
+  ros::NodeHandle nh_priv_{"~"};
+};
+
+::testing::AssertionResult expectOperationModeServiceCallResult
```
hslusarek, line 54: Docu missing

prbt_hardware_support/test/integrationtests/integrationtest_operation_mode.cpp
```diff
@@ -0,0 +1,216 @@
...
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+
+#include <thread>
+#include <memory>
+#include <chrono>
+#include <vector>
+#include <functional>
+#include <condition_variable>
+#include <atomic>
```
hslusarek: Not used -> remove.

prbt_hardware_support/test/integrationtests/integrationtest_operation_mode.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+
+#include <thread>
+#include <memory>
+#include <chrono>
+#include <vector>
+#include <functional>
+#include <condition_variable>
```
hslusarek: Not used -> remove.

prbt_hardware_support/test/integrationtests/integrationtest_operation_mode.cpp
```diff
@@ -0,0 +1,216 @@
...
+  modbus_server.terminate();
+  modbus_server_thread.join();
+}
+
+} // namespace prbt_hardware_support
+
+
+int main(int argc, char *argv[])
+{
+  ros::init(argc, argv, "integrationtest_brake_test_required");
```
hslusarek: Incorrect node name -> this is a integration test for the operation mode.

prbt_hardware_support/test/integrationtests/integrationtest_operation_mode.cpp
```diff
@@ -0,0 +1,216 @@
...
+/**
+ * @brief Send data via ModbusServerMock -> ModbusReadClient -> ModbusAdapterOperationMode connection
+ * and check that the expected result is returned via the service call.
+ *
+ * @note Due to the asynchronicity of the test each step of the sequence passed successful
+ *       allows the next step to be taken. See testing::AsyncTest for details.
+ *
+ * Test Sequence:
+ *    0. Start Modbus-server in separate thread. Make sure that the nodes are up.
+ *    1. Send a brake test required message with the correct API version.
```
hslusarek: Update docu

prbt_hardware_support/test/integrationtests/integrationtest_operation_mode.cpp
```diff
@@ -0,0 +1,216 @@
...
+ */
+
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+
+#include <thread>
+#include <memory>
+#include <chrono>
+#include <vector>
+#include <functional>
```
hslusarek: Not used -> remove.

prbt_hardware_support/test/integrationtests/integrationtest_operation_mode.cpp
```diff
@@ -0,0 +1,216 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+
+#include <thread>
+#include <memory>
+#include <chrono>
```
hslusarek: Not used -> remove.

prbt_hardware_support/test/integrationtests/integrationtest_operation_mode.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+#include <ros/ros.h>
+#include <std_srvs/SetBool.h>
+#include <std_srvs/Trigger.h>
+#include <std_msgs/Bool.h>
+
+#include <modbus/modbus.h>
+
+#include <prbt_hardware_support/ModbusMsgInStamped.h>
+#include <prbt_hardware_support/IsBrakeTestRequired.h>
```
hslusarek: Not used -> remove.

prbt_hardware_support/test/integrationtests/integrationtest_operation_mode.cpp
```diff
@@ -0,0 +1,216 @@
...
+#include <string>
+
+#include <ros/ros.h>
+#include <std_srvs/SetBool.h>
+#include <std_srvs/Trigger.h>
+#include <std_msgs/Bool.h>
+
+#include <modbus/modbus.h>
+
+#include <prbt_hardware_support/ModbusMsgInStamped.h>
```
hslusarek: Not used -> remove.

prbt_hardware_support/test/integrationtests/integrationtest_operation_mode.cpp
```diff
@@ -0,0 +1,216 @@
...
+  obj.run();
+}
+
+/**
+ * @brief OperationModeIntegrationTest checks if the chain
+ * ModbusServerMock -> ModbusReadClient -> ModbusAdapterOperationMode functions properly
+ *
+ * @note the test is derived from testing::AsyncTest which allows the asynchronous processes to re-sync
+ */
+class OperationModeIntegrationTest : public testing::Test, public testing::AsyncTest
```
hslusarek: `testing::AsyncTest` really needed? I do not see why.

prbt_hardware_support/test/integrationtests/integrationtest_operation_mode.cpp
```diff
@@ -0,0 +1,216 @@
...
+	prbt_hardware_support::GetOperationMode srv;
+
+  waitForNode("/pilz_modbus_read_client_node");
+  waitForNode("/modbus_adapter_operation_mode_node");
+
+  /**********
+   * Step 1 *
+   **********/
+  modbus_server.setHoldingRegister(std::vector<uint16_t> {MODBUS_API_VERSION_VALUE, 0, 0, 0, 0, 0},
+                                   index_of_first_register_to_read);
```
hslusarek: "Implicit conversion from `int` to `unsigned int`".

Please check **ALL** places where `index_of_first_register_to_read` is used and refactor the implicit conversions.

prbt_hardware_support/test/integrationtests/integrationtest_operation_mode.cpp
```diff
@@ -0,0 +1,216 @@
...
+#include <prbt_hardware_support/GetOperationMode.h>
+#include <prbt_hardware_support/OperationModes.h>
+#include <prbt_hardware_support/ros_test_helper.h>
+
+#include <pilz_testutils/async_test.h>
+
+namespace prbt_hardware_support
+{
+
+using ::testing::_;
```
hslusarek: Remove **ALL** `using ::testing:: ...` because nobody is using this stuff.

prbt_hardware_support/test/integrationtests/integrationtest_operation_mode.cpp
```diff
@@ -0,0 +1,216 @@
...
+   * Step 5 *
+   **********/
+  modbus_server.setHoldingRegister(std::vector<uint16_t>{MODBUS_API_VERSION_VALUE, 0, 0, 0, 0, 99},
+                                   index_of_first_register_to_read);
+	EXPECT_TRUE(expectOperationModeServiceCallResult(operation_mode_client, OperationModes::UNKNOWN, 10));
+
+  /**********
+   * Step 4 *
+   **********/
+  modbus_server.terminate();
```
hslusarek, line 174: What happens if an exception is thrown? I suggest moving the initialization and termination of the modbus thread to the `setUp()` and `tearDown()` functions.

prbt_hardware_support/test/integrationtests/integrationtest_operation_mode.cpp
```diff
@@ -0,0 +1,216 @@
...
+ *    4. Terminate ModbusServerMock.
+ *
+ * Expected Results:
+ *    0. -
+ *    1. A service call is successfull and returns a positive result.
+ *    2. A service call is successfull and returns a positive result.
+ *    3. A service call is successfull and returns a negative result.
+ *    4. -
+ */
+TEST_F(OperationModeIntegrationTest, testOperationModeRequestAnnouncement)
```
hslusarek, line 101: Why are all tests performed in one big test? This decreases readability and understandability for no good reason.

prbt_hardware_support/test/integrationtests/integrationtest_operation_mode.cpp
```diff
@@ -0,0 +1,216 @@
...
+#include <vector>
+#include <functional>
+#include <condition_variable>
+#include <atomic>
+#include <string>
+
+#include <ros/ros.h>
+#include <std_srvs/SetBool.h>
+#include <std_srvs/Trigger.h>
+#include <std_msgs/Bool.h>
```
hslusarek: Remove `#include <std_srvs/SetBool.h>`, `#include <std_srvs/Trigger.h>` and `#include <std_msgs/Bool.h>`

prbt_hardware_support/test/integrationtests/integrationtest_operation_mode.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+using ::testing::_;
+using ::testing::InSequence;
+using ::testing::Invoke;
+using ::testing::InvokeWithoutArgs;
+
+static constexpr uint16_t MODBUS_API_VERSION_VALUE {2};
+static const std::string SERVICE_OPERATION_MODE = "/prbt/get_operation_mode";
+
+template<class T>
```
hslusarek: No template needed. Use the concrete type.

prbt_hardware_support/test/integrationtests/integrationtest_operation_mode.test
```diff
@@ -0,0 +1,49 @@
...
+<!-- Modbus config -->
+<arg name="modbus_server_ip" default="127.0.0.1"/>
+<arg name="modbus_server_port" default="20701"/>
+<arg name="index_of_first_register_to_read" default="969" />
+<arg name="modbus_register_size" default="975" />
+<arg name="num_registers_to_read" default="6" />
+<arg name="safety_hw" default="pss4000" />
+
+<node name="pilz_modbus_read_client_node" pkg="prbt_hardware_support" type="pilz_modbus_read_client_node">
+  <param name="index_of_first_register_to_read" value="$(arg index_of_first_register_to_read)"/>
```
hslusarek, line 29: Remove because line exists twice. See some lines below.

prbt_hardware_support/test/integrationtests/integrationtest_operation_mode.cpp
```diff
@@ -0,0 +1,216 @@
...
+   * Setup *
+   **********/
+  std::string ip;
+  int port;
+  ASSERT_TRUE(nh_priv_.getParam("modbus_server_ip", ip));
+  ASSERT_TRUE(nh_priv_.getParam("modbus_server_port", port));
+
+  int modbus_register_size, num_registers_to_read, index_of_first_register_to_read;
+  ASSERT_TRUE(nh_priv_.getParam("modbus_register_size", modbus_register_size));
+  ASSERT_TRUE(nh_priv_.getParam("num_registers_to_read", num_registers_to_read));
```
hslusarek: Not used -> remove.

prbt_hardware_support/test/integrationtests/integrationtest_operation_mode.test
```diff
@@ -0,0 +1,49 @@
...
+<rosparam ns="api_spec" command="load" file="$(find prbt_hardware_support)/config/modbus_api_spec_$(arg safety_hw).yaml" />
+
+<node name="modbus_adapter_operation_mode_node" pkg="prbt_hardware_support" type="modbus_adapter_operation_mode_node">
+</node>
+
+<test test-name="integrationtest_operation_mode" pkg="prbt_hardware_support" type="integrationtest_operation_mode">
+  <param name="modbus_server_ip" value="$(arg modbus_server_ip)"/>
+  <param name="modbus_server_port" value="$(arg modbus_server_port)"/>
+  <param name="modbus_register_size" value="$(arg modbus_register_size)"/>
+  <param name="num_registers_to_read" value="$(arg num_registers_to_read)"/>
```
hslusarek: Not used -> remove.

prbt_hardware_support/test/integrationtests/integrationtest_stop1.test
```diff
@@ -30,6 +30,9 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 <arg name="modbus_register_size" default="514" />
 <arg name="num_registers_to_read" default="2" />
 
+<arg name="safety_hw" default="pnoz" />
+<rosparam ns="api_spec" command="load" file="$(find prbt_hardware_support)/config/modbus_api_spec_$(arg safety_hw).yaml" />
+
 <node name="pilz_modbus_read_client_node" pkg="prbt_hardware_support" type="pilz_modbus_read_client_node">
   <param name="index_of_first_register_to_read" value="$(arg index_of_first_register_to_read)"/>
```
hslusarek: Line exists twice -> remove.

hslusarek (changes requested):
Tests not completely reviewed yet.

prbt_hardware_support/include/prbt_hardware_support/modbus_api_definitions.h
agutenkunst: https://github.com/PilzDE/pilz_robots/issues/144

prbt_hardware_support/include/prbt_hardware_support/adapter_operation_mode.h
agutenkunst: Reference: http://blog.ycshao.com/2012/03/18/effective-c-item-7-declare-destructors-virtual-in-polymorphic-base-classes/

agutenkunst: Fixed.

prbt_hardware_support/include/prbt_hardware_support/adapter_operation_mode.h
agutenkunst: Fixed.

prbt_hardware_support/include/prbt_hardware_support/opertion_modes.h
agutenkunst: I really also dislike the part with this file. The intention is to seperate two concepts: ROS msg enum vs raw register values.

prbt_hardware_support/include/prbt_hardware_support/opertion_modes.h
agutenkunst: Fixed.

prbt_hardware_support/test/unittests/unittest_modbus_adapter_operation_mode.cpp
```diff
@@ -0,0 +1,251 @@
...
+ *
+ * Test Sequence:
+ *  1. Publish modbus message informing about changing operation mode. Repeat for all possible operation modes.
+ *
+ * Expected Results:
+
+  ros::Duration(OPERATION_MODE_CHANGE_WAIT_TIME_S).sleep();
+
+  prbt_hardware_support::GetOperationMode srv;
+  EXPECT_FALSE(operation_mode_client_.call(srv));
```
hslusarek: Are these three lines an copy error? Or do they have a purpose here?

prbt_hardware_support/src/adapter_operation_mode.cpp
agutenkunst: Done. ;-)

prbt_hardware_support/src/adapter_operation_mode.cpp
agutenkunst: Yep but I would keep it for now:
1. Consistent
2. Keep boost::optional away from the code base for now
3. A bit more expressiv for now

prbt_hardware_support/include/prbt_hardware_support/modbus_adapter_operation_mode.h
agutenkunst: The intention behind `AdapterOperationMode` is to have a base class for potential other derived classes working with other Protocols. Surely could be improved.

prbt_hardware_support/include/prbt_hardware_support/adapter_operation_mode.h
agutenkunst: See comment above

prbt_hardware_support/include/prbt_hardware_support/modbus_api_spec.h
agutenkunst: yep

prbt_hardware_support/include/prbt_hardware_support/modbus_msg_wrapper.h
agutenkunst: Not sure

prbt_hardware_support/include/prbt_hardware_support/modbus_msg_sto_wrapper.h
agutenkunst: done

prbt_hardware_support/include/prbt_hardware_support/operation_mode_filter.h
agutenkunst: Will keep for now

prbt_hardware_support/include/prbt_hardware_support/operation_mode_filter.h
agutenkunst: Will keep for now

prbt_hardware_support/include/prbt_hardware_support/modbus_msg_operation_mode_wrapper_exception.h
agutenkunst: Done

prbt_hardware_support/msg/OperationModes.msg
agutenkunst: Fixed

prbt_hardware_support/test/unittests/unittest_modbus_adapter_operation_mode.cpp
```diff
@@ -0,0 +1,251 @@
...
+ */
+TEST_F(ModbusAdapterOperationModeTest, testDisconnect)
+{
+  uint32_t offset{0};
+  std::vector<uint16_t> holding_register;
+  ModbusMsgInStampedPtr msg{createDefaultModbusMsgIn(offset, holding_register)};
+  msg->disconnect.data = true;
+  modbus_topic_pub_.publish(msg);
+
+  ASSERT_TRUE(waitForServiceCallFailure());
```
hslusarek: The service can never fail, see [here](https://github.com/PilzDE/pilz_robots/blob/f35d5173d761f86a19a9d9c88edfddc3b36090a3/prbt_hardware_support/src/adapter_operation_mode.cpp#L53). There is only one exception: The service fails, if it never has been initialized. In other words, this test only checks that the service never was initialized. 

In case of a disconnect message, I would expect that the service-call is successful but returns OperationMode==Unknown. This, however, implies that at first the tests ensures that the service is initialized and then sends a disconnect message.

prbt_hardware_support/test/unittests/unittest_modbus_adapter_operation_mode.cpp
```diff
@@ -0,0 +1,251 @@
...
+{
+  std::vector<uint16_t> holding_register;
+      modbus_topic_pub_.publish(createDefaultOpModeModbusMsg(
+      0,
+      MODBUS_API_VERSION_REQUIRED,
+      test_api_spec.getRegisterDefinition(modbus_api_spec::OPERATION_MODE),
+      test_api_spec.getRegisterDefinition(modbus_api_spec::VERSION))
+      );
+
+  ASSERT_TRUE(waitForServiceCallFailure());
```
hslusarek: In case of a unknown operation mode I expect that the service-call is successful but returns OperationMode==Unknown.

prbt_hardware_support/include/prbt_hardware_support/modbus_msg_operation_mode_wrapper.h
```diff
@@ -0,0 +1,95 @@
...
+   * @return true if a brake test required flag is defined, false otherwise.
+   */
+  bool hasOperationMode(const ModbusMsgInStampedConstPtr& modbus_msg_raw) const;
+};
+
+inline ModbusMsgOperationModeWrapper::ModbusMsgOperationModeWrapper(const ModbusMsgInStampedConstPtr& modbus_msg_raw,
+                                                            const ModbusApiSpec& api_spec):
+  ModbusMsgWrapper(modbus_msg_raw, api_spec)
+{
+  if(!hasVersion(msg_))
```
hslusarek: Check already exists in base class ModbusMsgWrapper. Therefore, i suggest to remove the check here.

prbt_hardware_support/test/integrationtests/integrationtest_operation_mode.cpp
agutenkunst: Really would like to have a modbus_server that is not member of the test

prbt_hardware_support/test/integrationtests/integrationtest_operation_mode.test
agutenkunst: Both the test node as well as the pilz_modbus_read_client_node need this line

prbt_hardware_support/test/unittests/unittest_modbus_adapter_operation_mode.cpp
```diff
@@ -0,0 +1,251 @@
...
+TEST_F(ModbusAdapterOperationModeTest, testModbusIncorrectApiVersion)
+{
+  std::vector<uint16_t> holding_register;
+
+  modbus_topic_pub_.publish(
+    createDefaultOpModeModbusMsg(
+      OPERATION_MODES.at(0),
+      MODBUS_API_VERSION_REQUIRED,
+      test_api_spec.getRegisterDefinition(modbus_api_spec::OPERATION_MODE),
+      0 /* incorrect version */)
```
hslusarek: This argument does not define the required API version but the register-index from which the API version is read. There actually exists no code which checks if the correct API version is used. So, first we need to add code which actually checks for the correct API version. Then this tests needs to be fixed because like said before: The service can never fail, see [here](https://github.com/PilzDE/pilz_robots/blob/f35d5173d761f86a19a9d9c88edfddc3b36090a3/prbt_hardware_support/src/adapter_operation_mode.cpp#L53). There is only one exception: The service fails, if it never has been initialized. In other words, this test only checks that the service never was initialized.

prbt_hardware_support/include/prbt_hardware_support/modbus_msg_wrapper.h
martiniil: I would keep the `virtual` for the destructor. All other functions are not intended for over-writing.

prbt_hardware_support/README.md
agutenkunst: I would opt for a dedicated document on all these safety related features along with the general concept rather than going into detail here and there...

prbt_hardware_support/test/unittests/unittest_modbus_adapter_operation_mode.cpp
```diff
@@ -0,0 +1,251 @@
...
+{
+  auto max_required_index = std::max(test_api_spec.getRegisterDefinition(modbus_api_spec::VERSION),
+                                     test_api_spec.getRegisterDefinition(modbus_api_spec::OPERATION_MODE));
+  auto msg{createDefaultOpModeModbusMsg(OPERATION_MODES.at(0),
+                                        MODBUS_API_VERSION_REQUIRED,
+                                        max_required_index - 1,
+                                        max_required_index - 1)};
+  modbus_topic_pub_.publish(msg);
+
+  ASSERT_TRUE(waitForServiceCallFailure());
```
hslusarek: The service can never fail, see [here](https://github.com/PilzDE/pilz_robots/blob/f35d5173d761f86a19a9d9c88edfddc3b36090a3/prbt_hardware_support/src/adapter_operation_mode.cpp#L53). There is only one exception: The service fails, if it never has been initialized. In other words, this test only checks that the service never was initialized. 

In my opinion, this test should check what happens in case a short register range is sent, when the service is already up and running.

prbt_hardware_support/test/unittests/unittest_modbus_adapter_operation_mode.cpp
```diff
@@ -0,0 +1,251 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+#include <algorithm>
+#include <functional>
+
+#include <std_msgs/Bool.h>
```
hslusarek: Not needed -> Remove.

prbt_hardware_support/test/unittests/unittest_modbus_adapter_operation_mode.cpp
```diff
@@ -0,0 +1,251 @@
...
+ * GNU Lesser General Public License for more details.
+
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+#include <algorithm>
+#include <functional>
```
hslusarek: Not needed -> Remove.

prbt_hardware_support/test/unittests/unittest_modbus_adapter_operation_mode.cpp
```diff
@@ -0,0 +1,251 @@
...
+static constexpr double OPERATION_MODE_CHANGE_WAIT_TIME_S{2.0};
+static const std::vector<unsigned int> OPERATION_MODES{1, 2, 3};
+
+/**
+ * @brief Test fixture for unit-tests of the ModbusAdapterOperationMode.
+ *
+ * Publish messages on the modbus topic and call the operation_mode service
+ * in order to check if the expectations are met.
+ */
+class ModbusAdapterOperationModeTest : public testing::Test, public testing::AsyncTest
```
hslusarek: This test class does not need to be derived from AsyncTest class.

prbt_hardware_support/test/unittests/unittest_modbus_adapter_operation_mode.cpp
```diff
@@ -0,0 +1,251 @@
...
+ *  1. Publish modbus message with holding registers not covering the expected range
+ *
+ * Expected Results:
+ *  1. The service call is not successful.
+ */
+TEST_F(ModbusAdapterOperationModeTest, testModbusWithShortRegisterRange)
+{
+  auto max_required_index = std::max(test_api_spec.getRegisterDefinition(modbus_api_spec::VERSION),
+                                     test_api_spec.getRegisterDefinition(modbus_api_spec::OPERATION_MODE));
+  auto msg{createDefaultOpModeModbusMsg(OPERATION_MODES.at(0),
```
hslusarek: Why not use the offical `OperationMode` enum values?

prbt_hardware_support/test/unittests/unittest_modbus_adapter_operation_mode.cpp
```diff
@@ -0,0 +1,251 @@
...
+    {
+      return ::testing::AssertionSuccess();
+    }
+    rate.sleep();
+  }
+
+  return ::testing::AssertionFailure() << "Reached timeout waiting for expected operation mode.";
+}
+
+bool ModbusAdapterOperationModeTest::waitForServiceCallFailure(double timeout)
```
hslusarek: Why does the method `waitForOperationMode` has `::testing::AssertionResult` as return value and this method has `bool` as return value?

prbt_hardware_support/test/unittests/unittest_modbus_adapter_operation_mode.cpp
```diff
@@ -0,0 +1,251 @@
...
+#include <algorithm>
+#include <functional>
+
+#include <std_msgs/Bool.h>
+
+#include <prbt_hardware_support/modbus_adapter_operation_mode.h>
+#include <prbt_hardware_support/modbus_topic_definitions.h>
+#include <prbt_hardware_support/modbus_msg_in_utils.h>
+
+#include <pilz_testutils/async_test.h>
```
hslusarek: Not  needed -> Remove.

prbt_hardware_support/test/unittests/unittest_modbus_adapter_operation_mode.cpp
```diff
@@ -0,0 +1,251 @@
...
+static constexpr int DEFAULT_QUEUE_SIZE_MODBUS{1};
+static const std::string SERVICE_NAME_OPERATION_MODE = "/prbt/get_operation_mode";
+
+static constexpr unsigned int MODBUS_API_VERSION_REQUIRED{2};
+
+static const ModbusApiSpec test_api_spec{ {modbus_api_spec::VERSION, 1},
+                                          {modbus_api_spec::OPERATION_MODE, 11} };
+
+static constexpr double OPERATION_MODE_CHANGE_WAIT_TIME_S{2.0};
+static const std::vector<unsigned int> OPERATION_MODES{1, 2, 3};
```
hslusarek, line 38: An `std::array` can be a `constexpr` because it does not uses dynamic memory allocation.

prbt_hardware_support/test/unittests/unittest_modbus_adapter_operation_mode.cpp
```diff
@@ -0,0 +1,251 @@
+/*
+ * Copyright (c) 2018 Pilz GmbH & Co. KG
```
hslusarek: Fix year -> 2019

prbt_hardware_support/test/unittests/unittest_modbus_adapter_operation_mode.cpp
```diff
@@ -0,0 +1,251 @@
...
+ * @brief Test fixture for unit-tests of the ModbusAdapterOperationMode.
+ *
+ * Publish messages on the modbus topic and call the operation_mode service
+ * in order to check if the expectations are met.
+ */
+class ModbusAdapterOperationModeTest : public testing::Test, public testing::AsyncTest
+{
+public:
+  ModbusAdapterOperationModeTest();
+  ~ModbusAdapterOperationModeTest();
```
hslusarek: Unnecessary -> Remove.

prbt_hardware_support/test/unittests/unittest_modbus_api_spec.cpp
```diff
@@ -0,0 +1,93 @@
+/*
+ * Copyright (c) 2018 Pilz GmbH & Co. KG
```
hslusarek: Fix year -> 2019.

prbt_hardware_support/include/prbt_hardware_support/opertion_modes.h
martiniil: @agutenkunst You have this separation here, even if you instead directly return the values from the ROS msg:
```
inline OperationMode ModbusMsgOperationModeWrapper::getOperationMode() const
{
switch(getRegister(api_spec_.getRegisterDefinition(modbus_api_spec::OPERATION_MODE))){
        case 0:
            return OperationMode::UNKNOWN;
```
I see it like hslusarek.
The only reason for keeping it is type safety.

agutenkunst: So keep it or leave it? @martiniil

prbt_hardware_support/test/unittests/unittest_modbus_api_spec.cpp
```diff
@@ -0,0 +1,93 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+#include <memory>
```
hslusarek: Not needed -> Remove.

prbt_hardware_support/include/prbt_hardware_support/operation_mode_filter.h
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+private:
+  void cb(const EventType &evt)
+  {
+    try
+    {
+      prbt_hardware_support::ModbusMsgOperationModeWrapper msg(evt.getMessage(), api_spec_);
+
+      if (!msg.isDisconnect())
```
hslusarek: Check already in constructor of base class of `ModbusMsgOperationModeWrapper`. Therefore, I suggest to remove this check.

prbt_hardware_support/test/unittests/unittest_operation_mode_filter.cpp
```diff
@@ -0,0 +1,163 @@
...
+{
+    public:
+        MOCK_METHOD1(modbusInMsgCallback, void(ModbusMsgInStampedConstPtr msg));
+};
+
+/**
+ * @brief Tests that the filter properly only passes message to the callback that have a operation mode
+ *        when starting with state true.
+ */
+TEST_F(OperationModeFilterTest, passOperationModeMsg)
```
hslusarek, line 73: The test docu states that the idea is to check that only messages are passed on if the messages contains an operation mode. Currently, however, the test is incomplete because it does not test the case in which the modbus-msg does not contain an operation mode. Merge this test with the test `messageWithOutOperationMode` because otherwise it makes no sense for the reader.

prbt_hardware_support/include/prbt_hardware_support/operation_mode_filter.h
```diff
@@ -0,0 +1,115 @@
...
+ *
+ * \code
+ *   message_filters::Subscriber sub<MsgStamped>(nh, TOPIC_NAME, 1);
+ *   message_filters::OperationMode filter<MsgStamped>(&sub));
+ *
+ *   // Register all callback receiving filtered output
+ *   filter.registerCallback(&filteredCallback);
+ * \endcode
+ */
+template <typename M>
```
hslusarek: Why do we need this template? We are dealing with Modbus-Msgs and we know this because in the `cb()` function it is even hard coded by the fact that the `ModbusMsgOperationModeWrapper` is used. Therefore, is suggest to remove the template here and derive the class from `SimpleFilter<ModbusMsgInStamped>` instead of `SimpleFilter<M>`.

prbt_hardware_support/test/unittests/unittest_operation_mode_filter.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+class OperationModeFilterTest : public testing::Test
+{
+    public:
+        MOCK_METHOD1(modbusInMsgCallback, void(ModbusMsgInStampedConstPtr msg));
+};
+
+/**
+ * @brief Tests that the filter properly only passes message to the callback that have a operation mode
+ *        when starting with state true.
```
hslusarek: "when starting with state true.", what does this mean?

prbt_hardware_support/test/unittests/unittest_operation_mode_filter.cpp
```diff
@@ -0,0 +1,163 @@
...
+    // Construct valid message
+    ModbusMsgInBuilder builder(test_api_spec);
+    builder.setApiVersion(MODBUS_API_VERSION_REQUIRED)
+           .setOperationMode(OperationMode::T1);
+
+    first_filter.signalMessage(builder.build());
+}
+
+/**
+ * @brief Tests that the filter properly only passes message to the callback that have a operation mode
```
hslusarek: Update docu.

prbt_hardware_support/test/unittests/unittest_operation_mode_filter.cpp
```diff
@@ -0,0 +1,163 @@
...
+using ::testing::Return;
+using ::testing::Throw;
+using namespace prbt_hardware_support;
+
+static const ModbusApiSpec test_api_spec{ {modbus_api_spec::VERSION, 696},
+                                          {"FakeValue", 511},
+                                          {modbus_api_spec::OPERATION_MODE,737} };
+
+
+ModbusMsgInStampedPtr createDefaultBrakeTestModbusMsg(bool brake_test_required)
```
hslusarek: Function not used -> Remove function.

hslusarek (changes requested):

prbt_hardware_support/test/unittests/unittest_modbus_adapter_operation_mode.cpp
agutenkunst: but std::initializer_list is not constexpr!

prbt_hardware_support/test/integrationtests/integrationtest_operation_mode.cpp
agutenkunst: This is a general issue with the tests. Even TearDown() won't get called if a uncaught exception is thrown...

martiniil (approved):

Issue #153, opened 2019-06-26
Set modbus server IP adress from command line


martiniil:
During startup via `moveit_planning_execution.launch` a default modbus server ip adress should be used depending on `sto:=pnoz` or `sto:=pss4000`. Otherwise, the user would want to set it via a command line argument.

PR #157, opened 2019-06-26
Make modbus server ip configurable from cmd line

martiniil:
- Targets #153
- Add default ip for pss4000


jschleicher:
The package [PilzDE/pilz_templates](https://github.com/pilzDE/pilz_templates) is meant to allow easy configuration of all parameters (including IP address etc). In my opinion an overly long command line shouldn't be the goal. Instead we should encourage the user to build and use their own launch files as described in the tutorials.

PR #160, opened 2019-06-28
Fix/auto modbus register range

### LINKED ISSUE #154: ModbusReadClient does not take into account register range from api spec
Ansatz:

Determine minimal modbus register range from api spec on parameter server.

martiniil:
Fixes #154

### DIFF
--- a/prbt_hardware_support/include/prbt_hardware_support/modbus_api_spec.h
+++ b/prbt_hardware_support/include/prbt_hardware_support/modbus_api_spec.h
@@ -18,9 +18,11 @@
 #ifndef MODBUS_API_SPEC_H
 #define MODBUS_API_SPEC_H
 
+#include <algorithm>
+#include <initializer_list>
 #include <map>
 #include <string>
-#include <initializer_list>
+#include <utility>
 
 #include <ros/ros.h>
 #include <xmlrpcpp/XmlRpc.h>
@@ -114,6 +116,33 @@ public:
     {
       throw ModbusApiSpecException(e.what());
     }
+  }
+  inline unsigned int getMinRegisterDefinition() const
+  {
+    if (register_mapping_.empty())
+    {
+      throw ModbusApiSpecException("Cannot read values. Api spec is empty.");
+    }
+
+    typedef std::pair<std::string, unsigned int> RegisterMappingEntry;
+    auto it = std::min_element(register_mapping_.begin(), register_mapping_.end(),
+                               [] (const RegisterMappingEntry& a, const RegisterMappingEntry& b)
+                               { return a.second < b.second; });
+    return it->second;
+  }
+
+  inline unsigned int getMaxRegisterDefinition() const
+  {
+    if (register_mapping_.empty())
+    {
+      throw ModbusApiSpecException("Cannot read values. Api spec is empty.");
+    }
+
+    typedef std::pair<std::string, unsigned int> RegisterMappingEntry;
+    auto it = std::max_element(register_mapping_.begin(), register_mapping_.end(),
+                               [] (const RegisterMappingEntry& a, const RegisterMappingEntry& b)
+                               { return a.second < b.second; });
+    return it->second;
   }
 
 private:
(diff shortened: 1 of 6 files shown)


martiniil:
I can not get rid of this gap in function coverage:

> prbt_hardware_support::ModbusApiSpecTemplated<ros::NodeHandle>::getMinMaxRegisterDefintion() const::{lambda(std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, unsigned int> const&, std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, unsigned int> const&)#1}::_FUN(std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, unsigned int> const&, std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, unsigned int> const&)

What I know is that it is a static invoke function that is generated from the lambda expression in `getMinMaxRegisterDefintion()`. I assume this is a general issue when using lambdas.

jschleicher:
What gcc version are you using? In 8.3.0 the following patch was applied, that may fix this problem (if that is the case, it is not your fault for sure and you can ignore the line):
https://patchwork.ozlabs.org/patch/968991/
https://github.com/gcc-mirror/gcc/commit/7de708f8f52a6e674eeb9ba729df2682bae34aea#diff-e4f19c81816dc138788262149d2fa68f

martiniil:
@jschleicher Thank you! I can confirm that the problem is fixed using gcc 8.3.0 and lcov 1.14

prbt_hardware_support/src/pilz_modbus_read_client_node.cpp
```diff
@@ -39,42 +41,71 @@ int main(int argc, char **argv)
 {
   ros::init(argc, argv, "modbus_read_node");
 
-  ros::NodeHandle nh {"~"};
+  ros::NodeHandle pnh{"~"};
+  ros::NodeHandle nh;
+
+  bool has_register_range_parameters = pnh.hasParam(PARAM_NUM_REGISTERS_TO_READ_STR) &&
```
agutenkunst: Seems to me that you could condense all parts handling this into one section rather than distributing it all over the file

prbt_hardware_support/src/pilz_modbus_read_client_node.cpp
```diff
@@ -39,42 +41,71 @@ int main(int argc, char **argv)
...
-  nh.param<std::string>(PARAM_MODBUS_TOPIC_NAME_STR, modbus_topic_name,
-                        TOPIC_MODBUS_READ);
-
+  pnh.param<std::string>(PARAM_MODBUS_TOPIC_NAME_STR, modbus_topic_name,
+                         TOPIC_MODBUS_READ);
 
   // LCOV_EXCL_STOP
 
-  prbt_hardware_support::PilzModbusReadClient modbus_client(nh,
+  if (!has_register_range_parameters)
```
agutenkunst: See comment above

prbt_hardware_support/include/prbt_hardware_support/modbus_api_spec.h
```diff
@@ -116,6 +118,20 @@ class ModbusApiSpecTemplated
     }
   }
 
+  inline std::pair<unsigned int, unsigned int> getMinMaxRegisterDefintion() const
```
agutenkunst: I would like to say that a comment is missing but somehow I feel that it rather a missing concept of something like "range". Maybe a dedicated class for more expressiveness?

agutenkunst: Or... just provide a getMinRegister() and a getMaxRegister() to KISS

agutenkunst (changes requested):

prbt_hardware_support/include/prbt_hardware_support/modbus_api_spec.h
agutenkunst: Why not just getMinRegister() does "Definition" provide any more useful information?

martiniil: It is consistent with the other methods: `setRegisterDefinition()` and so on.

martiniil: Did you choose it, because "setRegister()" would have been misleading?

agutenkunst: Yep. But also not ideal

agutenkunst (approved):

PR #145, opened 2019-06-17
Increase gcc checks

agutenkunst:
Add -Wconversions -pendantic to prbt_hardware_support

Needs 
https://github.com/PilzDE/pilz_robots/pull/162 :heavy_check_mark:
~~https://github.com/PilzDE/pilz_robots/pull/163~~
https://github.com/PilzDE/pilz_robots/pull/164 :heavy_check_mark:
https://github.com/PilzDE/pilz_robots/pull/165 :heavy_check_mark:
https://github.com/PilzDE/pilz_robots/pull/171 :heavy_check_mark:
https://github.com/PilzDE/pilz_robots/pull/174 :heavy_check_mark:

to be closed

PR #163, opened 2019-07-03
Catch possible overflow

agutenkunst:

### DIFF
--- a/prbt_hardware_support/include/prbt_hardware_support/modbus_msg_in_utils.h
+++ b/prbt_hardware_support/include/prbt_hardware_support/modbus_msg_in_utils.h
@@ -24,6 +24,8 @@
 #include <prbt_hardware_support/modbus_api_spec.h>
 #include <prbt_hardware_support/OperationModes.h>
 
+#include <boost/numeric/conversion/cast.hpp>
+
 namespace prbt_hardware_support
 {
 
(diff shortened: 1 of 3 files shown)


prbt_hardware_support/include/prbt_hardware_support/modbus_msg_in_utils.h
```diff
@@ -123,11 +125,20 @@ class ModbusMsgInBuilder
...
-      msg->holding_registers.layout.dim[0].size = tab_reg.size();
+      try
+      {
+        msg->holding_registers.layout.dim[0].size
+                                      = boost::numeric_cast<std_msgs::MultiArrayDimension::_size_type>(tab_reg.size());
+        msg->holding_registers.data = tab_reg;
+      }
+      catch(boost::numeric::positive_overflow& e)
+      {
+        ROS_ERROR("The modbus register is to big!");
```
ct2034, line 136: should this also fail?

ct2034: throw ModbusExceptionDisconnect ?

ct2034 (changes requested):

prbt_hardware_support/include/prbt_hardware_support/modbus_msg_in_utils.h
jschleicher: I'd suggest:
1. **use a `static_cast<>`**. Since you currently do not use the overflow but simply proceed, you could as good use the standard cast.
OR
2. **Not catch** Aborting the program gives the best possibilities to debug and most likely will issue bug reports, if someone breaks this part of the code in the future while ROS_ERRORs are silently ignored by most users. The error thrown gives more details about the overflow than any custom ModbusExceptionDisconnect.

ct2034:
@agutenkunst Please solve the conflicts. I don't blick it

agutenkunst:
Irrelevant by now due to 924dedbe6237aa444e367e37fce5187be7fd3e91

Closing.

PR #164, opened 2019-07-03
Fix type conversion by makeing timeout long

agutenkunst:


ct2034 (changes requested):
Please add a pointer to where this method is used

agutenkunst:
It is actually about the return value of getResponseTimeout()

https://github.com/PilzDE/pilz_robots/blob/a5f2b244923747dcaebf40ceeb75fcae5aa7a50d/prbt_hardware_support/src/libmodbus_client.cpp#L60

ct2034 (commented):
but who uses it?

ct2034 (approved):
ok :+1:

PR #168, opened 2019-07-04
Add default modbus server ip for pss4000

martiniil:
pnoz remains default, only choose other ip if `sto:=pss4000` in robot.launch


agutenkunst (approved):

jschleicher:
:+1 Thanks!
Just for the record: implements #153 and replaces #157.

PR #170, opened 2019-07-05
buildfarm badges

ct2034:
Adding badges to monitor buildfarm status in readme


agutenkunst (changes requested):
I would write Buildfarm instead of BF

Or we could go the extra mile to have it like MoveIt! https://github.com/ros-planning/moveit

ct2034:
please also review https://github.com/PilzDE/pilz_industrial_motion/pull/183

ct2034:
> Or we could go the extra mile to have it like MoveIt! https://github.com/ros-planning/moveit

no. We would need a script for that. And for the script we would need a pbi-item

agutenkunst:
OK.

agutenkunst (approved):

agutenkunst (approved):

PR #171, opened 2019-07-08
Fix type in multiplication

agutenkunst:

### DIFF
--- a/prbt_hardware_support/src/libmodbus_client.cpp
+++ b/prbt_hardware_support/src/libmodbus_client.cpp
@@ -59,7 +59,7 @@ unsigned long LibModbusClient::getResponseTimeoutInMs()
 {
   struct timeval response_timeout;
   modbus_get_response_timeout(modbus_connection_, &response_timeout);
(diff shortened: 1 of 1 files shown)


prbt_hardware_support/src/libmodbus_client.cpp
```diff
@@ -59,7 +59,7 @@ unsigned long LibModbusClient::getResponseTimeoutInMs()
 {
   struct timeval response_timeout;
   modbus_get_response_timeout(modbus_connection_, &response_timeout);
-  return response_timeout.tv_sec * 1000 + (response_timeout.tv_usec  / 1000);
+  return response_timeout.tv_sec * 1000l + (response_timeout.tv_usec  / 1000l);
```
ct2034: I misread it as a 1 the first time and had no idea what you were trying to do. ;-) ->
Can you use a capital L? I also think this might be more common

ct2034 (changes requested):

prbt_hardware_support/src/libmodbus_client.cpp
agutenkunst: Done

ct2034 (approved):

PR #175, opened 2019-07-09
Fix/delayed sto

### LINKED ISSUE #151: Hold service waits until recover returns
### Commit
836eeb30d93de4bd873a61373753bf4e61c163ab

### Steps to reproduce
How can the issue/bug be reproduced?
1. `roslaunch prbt_moveit_config moveit_planning_execution.launch sim:=false has_braketest_support:=true has_operation_mode_support:=true pipeline:=pilz_command_planner`
2. Press and immediately release enabling switch

### Expected behavior
Hold service is called as soon as it is detected via modbus.

### Observed behavior
Call of `/hold` and subsequent `/halt` service is delayed by more than 50ms.

### Logfiles
```
[DEBUG] [1561374871.068122937]: Sending new ROS-message.
[DEBUG] [1561374871.069153026]: Calling Unhold (Service: /prbt/manipulator_joint_trajectory_controller/unhold)
[DEBUG] [1561374871.070522761]: Calling Recover (Service: /prbt/driver/recover)
[ INFO] [1561374871.071264156]: Recovering XXX
[DEBUG] [1561374871.088116191]: Sending new ROS-message.
EMCY: 83#0000000000000000
EMCY: 84#0000000000000000
EMCY: 85#0000000000000000
EMCY: 86#0000000000000000
EMCY: 87#0000000000000000
EMCY: 88#0000000000000000
[ INFO] [1561374871.156928772]: Performing STOP due to STO signal
[DEBUG] [1561374871.156962179]: Calling Hold on controller (Service: /prbt/manipulator_joint_trajectory_controller/hold)
[DEBUG] [1561374871.358027119]: Calling Halt on driver (Service: /prbt/driver/halt)
[ INFO] [1561374871.359182521]: Halting down XXX
```

The delay between "Sending new ROS message" and "Performing STOP" is about 68 milliseconds in this log.

The delay triggers bug #93.

Probably the reason is the blocking `/recover` call. Is it possible to abort that or overrule it by a multithreaded `/halt` call?

martiniil:
fix #151


prbt_hardware_support/include/prbt_hardware_support/service_client_factory.h
```diff
@@ -0,0 +1,48 @@
...
+  /**
+   * @brief Waits for a service to exist and creates a ServiceClient.
+   *
+   * @param name Name of the service, to be resolved in the node namespace
+   * @return The ServiceClient
+   */
+  template <typename Service>
+  inline static ros::ServiceClient create(const std::string &name)
+  {
+    waitForService(name);
```
jschleicher: better add `while(ros::ok) { ROS_INFO_THROTTLED...}` as in http://wiki.ros.org/roscpp/Overview/Logging to give feedback to the user while waiting.
see https://github.com/PilzDE/pilz_robots/blob/10b4fb75ee28a35107087a024a174cd67026305f/prbt_hardware_support/include/prbt_hardware_support/wait_for_service.h#L38-L43

jschleicher (commented):
Where is your state diagram gone? Didn't you have an image in a previous version of this branch?

prbt_hardware_support/include/prbt_hardware_support/service_client_factory.h
martiniil: this is exactly the waitForService() that is called ...

jschleicher: okay. To avoid further confusion I'd suggest to unify both. Either keep your Factory class and refactor  braketest accordingly or stick to Hagen's functional approach and remove your class. Why would you need two approaches (files, tests, documentation) for the same functionality?
https://github.com/PilzDE/pilz_robots/blob/8bd4e2042462b7dc2768748bded6e2e150d234cb/prbt_hardware_support/src/brake_test_executor.cpp#L49-L58

martiniil: I agree. I would suggest to do the refactoring in a separate PR and concentrate on the STO fix here.

martiniil: Also I think the confusion is between ros::service::waitForService() and prbt_hardware_support::waitForService(), which can be eliminated by refactoring.

pilz_testutils/include/pilz_testutils/async_test.h
```diff
@@ -88,6 +88,11 @@ class AsyncTest
      */
     void barricade(std::initializer_list<std::string> clear_events);
 
+    /**
```
agutenkunst: Not used please drop.

pilz_testutils/src/async_test.cpp
```diff
@@ -52,4 +55,23 @@ void AsyncTest::triggerClearEvent(std::string event)
   cv_.notify_one();
 }
 
+void AsyncTest::printAsyncEvents()
```
agutenkunst: As long not used drop it please

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
```diff
@@ -100,9 +143,201 @@ class AdapterSto
...
+template <class T>
+void AdapterStoTemplated<T>::updateSto(const bool STO)
+{
+  std::lock_guard<std::mutex> lock(action_mutex_);
+
+  if (STO)
+  {
+    if (next_action_ == Action::HALTING)
+    {
+      next_action_ = Action::RECOVERING;
```
jschleicher: The transition HALTING->RECOVERING isn't supported.
Please also correct diag_state_sto_executor_action.png

prbt_hardware_support/test/include/prbt_hardware_support/service_client_mock.h
```diff
@@ -0,0 +1,92 @@
...
+namespace prbt_hardware_support_tests
+{
+
+/**
+ * @brief Mock for a ros::ServiceClient, calls are passed through to a given callback (which can be a mock method).
+ *
+ * @tparam S service type
+ */
+template <typename S>
+class ServiceClientMock
```
jschleicher: Is the ServiceClientMock class useful in more general and could be placed inside test_utils?

jschleicher (commented):
Thanks @martiniil for pushing this. I didn't look at the unittest. Just two inline comments:

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
```diff
@@ -55,43 +86,55 @@ class AdapterSto
...
-   * @brief Update stored STO value and react to STO changes.
-   * Reactions:
-   *  - STO == false:   Perform Stop 1
-   *  - STO == true:    unhold Controller + recover Drives
+   * @brief Depending on the latest action and the \p STO determine the next action.
    */
   void updateSto(const bool sto);
 
 private:
+  enum Action
```
agutenkunst: `enum class` maybe?

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
```diff
@@ -100,9 +143,201 @@ class AdapterSto
...
+template <class T>
+void AdapterStoTemplated<T>::run()
+{
+  Action current_action{Action::HALTING};
+  bool execute_action{false};
+
+  ros::Rate rate(10);
+  while (!terminate_)
+  {
+    {
```
agutenkunst, line 200: This part is hard to understand. As far as I understand you only need to lock the assignment.

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
```diff
@@ -100,9 +143,201 @@ class AdapterSto
...
+      std::lock_guard<std::mutex> lock(action_mutex_);
```
agutenkunst: Are you really sure you need this mutex. The enum should be atomic and you have on thread writting it right?

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
```diff
@@ -100,9 +143,201 @@ class AdapterSto
...
+  terminate();
+}
+
+template <class T>
+void AdapterStoTemplated<T>::runAsync()
+{
+  run_thread_ = std::thread{[this] { this->run(); }};
+}
+
+template <class T>
```
agutenkunst, line 222: Please provide a detailed description here

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
```diff
@@ -40,12 +42,41 @@ namespace prbt_hardware_support
  * specifies the time between holding the controller and halting the driver.
  * This allows the controller to perform a smooth stop before a driver disable
  * enables the breaks.
+ *
+ * @remark this class is templated for easier mocking. However for usability
+ * it can be used by AdapterSto
  */
-class AdapterSto
+template <class T = ros::ServiceClient>
+class AdapterStoTemplated
```
agutenkunst, line 57: Please add move detail to the class description on what you are trying to achieve

prbt_hardware_support/doc/diag_state_sto_executor_action.uxf
```diff
@@ -0,0 +1,195 @@
+<?xml version="1.0" encoding="UTF-8"?><diagram program="umlet" version="13.3">
```
agutenkunst: In your diagram as far as I understand some guard conditions could provide more insight?

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
```diff
@@ -100,9 +143,201 @@ class AdapterSto
...
+    run_thread_.join();
+  }
+}
+
+template <class T>
+void AdapterStoTemplated<T>::autoUpdate(Action finished_action, bool exec_result)
+{
+  std::lock_guard<std::mutex> lock(action_mutex_);
+
+  if (next_action_ == finished_action)
```
agutenkunst: Not readable.

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
```diff
@@ -100,9 +143,201 @@ class AdapterSto
...
+  {
+    switch (finished_action)
```
agutenkunst: Maybe some StateMachine Framework would help keeping this sane?

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
```diff
@@ -100,9 +143,201 @@ class AdapterSto
...
+    }
+  }
+}
+
+template <class T>
+bool AdapterStoTemplated<T>::execute(Action action)
+{
+  bool success{false};
+
+  switch (action)
```
agutenkunst: Any idea on some pattern not relying on switch cases? As soon as some other ACTION is introduced these will grow and spread...

agutenkunst: https://stackoverflow.com/a/126475

prbt_hardware_support/test/integrationtests/integrationtest_modbus_adapter_sto.cpp
```diff
@@ -211,33 +212,36 @@ TEST_F(ModbusAdapterStoTest, testSetupNoDisableService)
...
+ * Expected: Constructor finishes successfully without unhold service
  */
 TEST_F(ModbusAdapterStoTest, testSetupNoUnholdService)
 {
   manipulator_.advertiseHoldService(nh_, HOLD_SERVICE_T);
   manipulator_.advertiseHaltService(nh_, HALT_SERVICE_T);
   manipulator_.advertiseRecoverService(nh_, RECOVER_SERVICE_T);
 
-  modbus_sto_adapter_.reset(new ModbusAdapterSto(nh_, test_api_spec));
+  auto t = asyncConstructor();
```
agutenkunst: variabel naming! Clean Code.

prbt_hardware_support/test/integrationtests/integrationtest_modbus_adapter_sto.cpp
```diff
@@ -249,9 +253,12 @@ TEST_F(ModbusAdapterStoTest, testSetupNoRecoverService)
   manipulator_.advertiseUnholdService(nh_, UNHOLD_SERVICE_T);
   manipulator_.advertiseHaltService(nh_, HALT_SERVICE_T);
 
-  modbus_sto_adapter_.reset(new ModbusAdapterSto(nh_, test_api_spec));
+  auto t = asyncConstructor();
```
agutenkunst: See above

prbt_hardware_support/test/integrationtests/integrationtest_service_client_factory.cpp
```diff
@@ -0,0 +1,67 @@
...
+  return true;
+}
+
+/**
+ * @brief Tests that an asynchronous call to ServiceClientFactory::create waits for
+ *        the service to come up and returns a service client.
+ */
+TEST(ServiceClientFactoryIntegrationTest, testCreate)
+{
+  ros::AsyncSpinner spinner {1};
```
agutenkunst: Are you sure you need this?

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
martiniil: What do you mean? This triggers recover when STO false -> true.

agutenkunst (changes requested):

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
jschleicher: Exactly. But the shortcut is only desired for RECOVERING->HALTING. Once STO goes to false, the full halt is always executed (see doors specification of PSS4000 function block).

martiniil: Sorry, if the code is misleading, I am rewriting this part at the moment and hope to resolve the ambiguities.

The full halt (when STO goes false) is UNHOLDING -> HOLDING -> HALTING

HALTING -> RECOVERING for STO true is no shortcut

jschleicher: Or: Do you deliberately deviate from the PSS4000 spec and leave that "full stop" to the safety controller? As said: In my opinion the diag_state_sto_executor_action.png should be asymmetric: Only having one shortcut for RECOVERING->HALTING. In state HOLDING, the sto=true could be visualized as self-transition / effectively a no-op.

martiniil: Ok, I got it. Thanks!

agutenkunst:
I checked with b5bce3c5f2f4ccf3806233d1234bd7d96b552dd7

```
[DEBUG] [WallTime: 1563793840.023446066 /pilz_robots/prbt_hardware_support/src/pilz_modbus_client.cpp:161]: Sending new ROS-message.
Warning: recursive print statement has occurred.  Throwing out recursive print.
[ INFO] [WallTime: 1563793840.028309680 /ros_canopen/canopen_chain_node/src/ros_chain.cpp:132]: Recovering XXX
EMCY: 83#0000000000000000
EMCY: 84#0000000000000000
EMCY: 85#0000000000000000
EMCY: 86#0000000000000000
EMCY: 87#0000000000000000
EMCY: 88#0000000000000000
[DEBUG] [WallTime: 1563793840.267283104 /pilz_robots/prbt_hardware_support/src/pilz_modbus_client.cpp:161]: Sending new ROS-message.
Warning: recursive print statement has occurred.  Throwing out recursive print.
[ INFO] [WallTime: 1563793840.267619733 /pilz_robots/prbt_hardware_support/src/adapter_sto.cpp:68]: Performing STOP due to STO signal
[ INFO] [WallTime: 1563793840.470931269 /ros_canopen/canopen_chain_node/src/ros_chain.cpp:196]: Halting down XXX
```
so the time here would be ~45ms

I did a rebase against https://github.com/PilzDE/pilz_robots/pull/184 and got
```
[DEBUG] [WallTime: 1563797134.323560577 /pilz_robots/prbt_hardware_support/src/pilz_modbus_client.cpp:161]: Sending new ROS-message.
Warning: recursive print statement has occurred.  Throwing out recursive print.
[ERROR] [WallTime: 1563797134.323835277 /pilz_robots/prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h:166]: UPDATING IN THREAD 139794322971840
[ERROR] [WallTime: 1563797134.323937553 /pilz_robots/prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h:196]: Calling Recover (Service: /prbt/driver/recover)
[ INFO] [WallTime: 1563797134.325399648 /ros_canopen/canopen_chain_node/src/ros_chain.cpp:132]: Recovering XXX
[DEBUG] [WallTime: 1563797134.383626879 /pilz_robots/prbt_hardware_support/src/pilz_modbus_client.cpp:161]: Sending new ROS-message.
Warning: recursive print statement has occurred.  Throwing out recursive print.
[ERROR] [WallTime: 1563797134.383920953 /pilz_robots/prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h:166]: UPDATING IN THREAD 139794322971840
[ERROR] [WallTime: 1563797134.383966795 /pilz_robots/prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h:229]: Calling Hold on controller (Service: /prbt/manipulator_joint_trajectory_controller/hold)
[ WARN] [WallTime: 1563797134.385457115 /pilz_robots/pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h:99]: Trying to hold controller which is in hold mode already!
[ERROR] [WallTime: 1563797134.386955541 /pilz_robots/prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h:252]: Calling Halt on driver (Service: /prbt/driver/halt)
[ INFO] [WallTime: 1563797134.388161185 /ros_canopen/canopen_chain_node/src/ros_chain.cpp:196]: Halting down XXX
EMCY: 83#0000000000000000
EMCY: 84#0000000000000000
EMCY: 85#0000000000000000
EMCY: 86#0000000000000000
EMCY: 87#0000000000000000
EMCY: 88#0000000000000000
[ERROR] [WallTime: 1563797134.409911074 /pilz_robots/prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h:198]: Finished Recover (Service: /prbt/driver/recover)
[ WARN] [WallTime: 1563797134.409966943 /pilz_robots/prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h:208]: Lost STO Clearance during recover. Keeping controller in hold mode!
```

which should be what we want. But I can also produce a
```
[DEBUG] [WallTime: 1563797412.838820902 /pilz_robots/prbt_hardware_support/src/pilz_modbus_client.cpp:161]: Sending new ROS-message.
Warning: recursive print statement has occurred.  Throwing out recursive print.
[ERROR] [WallTime: 1563797412.839130300 /pilz_robots/prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h:166]: UPDATING IN THREAD 139743469784256
[ERROR] [WallTime: 1563797412.839230633 /pilz_robots/prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h:196]: Calling Recover (Service: /prbt/driver/recover)
[ INFO] [WallTime: 1563797412.840668482 /ros_canopen/canopen_chain_node/src/ros_chain.cpp:132]: Recovering XXX
EMCY: 83#0000000000000000
EMCY: 84#0000000000000000
EMCY: 85#0000000000000000
EMCY: 86#0000000000000000
EMCY: 87#0000000000000000
[DEBUG] [WallTime: 1563797412.926921287 /pilz_robots/prbt_hardware_support/src/pilz_modbus_client.cpp:161]: Sending new ROS-message.
Warning: recursive print statement has occurred.  Throwing out recursive print.
[ERROR] [WallTime: 1563797412.927224960 /pilz_robots/prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h:166]: UPDATING IN THREAD 139743469784256
[ERROR] [WallTime: 1563797412.927273668 /pilz_robots/prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h:229]: Calling Hold on controller (Service: /prbt/manipulator_joint_trajectory_controller/hold)
[ WARN] [WallTime: 1563797412.928813877 /pilz_robots/pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h:99]: Trying to hold controller which is in hold mode already!
EMCY: 88#0000000000000000
[ERROR] [WallTime: 1563797412.929511494 /pilz_robots/prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h:198]: Finished Recover (Service: /prbt/driver/recover)
[ WARN] [WallTime: 1563797412.929553917 /pilz_robots/prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h:208]: Lost STO Clearance during recover. Keeping controller in hold mode!
[ WARN] [WallTime: 1563797412.930461204 /pilz_robots/prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h:246]: Waiting before halt since trajectory is executing!
[ WARN] [WallTime: 1563797413.130625230 /pilz_robots/prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h:248]: Waiting done.
[ERROR] [WallTime: 1563797413.130678978 /pilz_robots/prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h:252]: Calling Halt on driver (Service: /prbt/driver/halt)
[ INFO] [WallTime: 1563797413.132201856 /ros_canopen/canopen_chain_node/src/ros_chain.cpp:196]: Halting down XXX
```
which seems to indicate that we expect the controller to be executing introducing a 20ms extra wait. Not sure where this is coming from...

prbt_hardware_support/test/include/prbt_hardware_support/service_client_mock.h
martiniil: yes!

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
martiniil: outdated.

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
martiniil: outdated.

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
martiniil: outdated.

prbt_hardware_support/doc/diag_state_sto_executor_action.uxf
martiniil: outdated.

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
martiniil: outdated.

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
martiniil: yes!

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
martiniil: done.

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
martiniil: done.

prbt_hardware_support/test/integrationtests/integrationtest_modbus_adapter_sto.cpp
martiniil: done.

prbt_hardware_support/test/integrationtests/integrationtest_modbus_adapter_sto.cpp
martiniil: Done

prbt_hardware_support/test/integrationtests/integrationtest_service_client_factory.cpp
martiniil: I think it is not needed.

martiniil:
I ran tests on both robots and compared b2a5c43 with 4ba39238983e42cc630628247cace5d9549a3381 (melodic-devel) 
The crucial part is the time between

> Sending new ROS-Message.

and 

> Halting down XXX

The outcome is as expected: Old (without fix): ~269ms, new: ~54ms

Old:
melodic_devel

New:
fix_delayed_sto

prbt_hardware_support/test/integrationtests/integrationtest_stop1.test
```diff
@@ -45,6 +46,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
   <param name="halt_service_name" value="$(arg halt_service_name)"/>
   <param name="hold_service_name" value="$(arg hold_service_name)"/>
   <param name="unhold_service_name" value="$(arg unhold_service_name)"/>
+  <param name="is_executing_service_name" value="$(arg is_executing_service_name)"/>
```
jschleicher: Why do you prefer parameters for specifying service names over [remapping](http://wiki.ros.org/roslaunch/XML/remap)?

prbt_hardware_support/include/prbt_hardware_support/service_client_factory.h
jschleicher: Is there already a PR for `waitForService` refactoring available? Or an open issue?

jschleicher: What about moving service_client_factory into a central package for easy re-usability?

prbt_hardware_support/include/prbt_hardware_support/sto_state_machine.h
```diff
@@ -0,0 +1,342 @@
...
+  /////////////////
+  // Transitions //
+  /////////////////
+
+  struct transition_table : mpl::vector<
+  //  Start                       Event          Target                      Action         Guard
+  // +---------------------------+--------------+---------------------------+--------------+----------+
+  Row< RobotInactive             , sto_updated  , Enabling                  , recover_start, sto_true >,
+  Row< RobotInactive             , sto_updated  , none                      , none         , sto_false>,
+  Row< Enabling                  , sto_updated  , none                      , none         , sto_true >,
```
jschleicher, line 338: Do you need those empty self-transitions with no action?

prbt_hardware_support/test/include/prbt_hardware_support/pilz_manipulator_mock.h
```diff
@@ -39,70 +39,131 @@ class ManipulatorMock
     void advertiseUnholdService(ros::NodeHandle nh, std::string unhold_service_name);
     void advertiseHaltService(ros::NodeHandle nh, std::string halt_service_name);
     void advertiseRecoverService(ros::NodeHandle nh, std::string recover_service_name);
+    void advertiseIsExecutingService(ros::NodeHandle nh, std::string is_executing_service_name);
```
jschleicher: Imho a better name would be `PilzControllerMock` since it mocks the ros_control-derived `PilzJointTrajectoryController`

prbt_hardware_support/test/integrationtests/integrationtest_modbus_adapter_sto.cpp
```diff
@@ -318,46 +340,125 @@ TEST_F(ModbusAdapterStoTest, testRemoveUnholdService)
...
  * Controller must be holded and driver must be disabled.
  *
- * @tests{Hold_driver_if_STO_false,
- *  Tests that a message giving sto request leads to driver halt call.
- * }
- *
- * @tests{No_new_commands_during_STO_false,
- *  Tests that a message giving sto request leads to controller hold call.
+ * @tests{Stop1_Trigger,
+ *  Test that Stop 1 is triggered if STO value changes to false.
```
jschleicher, line 389: Seems like a Duplication of the unittest `TEST_F(AdapterStoTest, testEnableStopEnable)` and Integrationtest `TEST_F(Stop1IntegrationTest, testServiceCallbacks)`?

jschleicher (dismissed):
The arrows in prbt_hardware_support/doc/architecture_overview.png are overlapping:
grafik
Looks good overall, please see some inline comments (maybe resolve them into refactoring-issues).

martiniil:
Related issue: https://github.com/PilzDE/pilz_robots/issues/206

prbt_hardware_support/test/integrationtests/integrationtest_stop1.test
martiniil: See #208

prbt_hardware_support/include/prbt_hardware_support/sto_state_machine.h
martiniil: Yes, otherwise no_transition is called.
https://www.boost.org/doc/libs/1_64_0/libs/msm/doc/HTML/ch03s02.html#d0e471

prbt_hardware_support/test/integrationtests/integrationtest_modbus_adapter_sto.cpp
martiniil: These are tests on different levels.

prbt_hardware_support/test/include/prbt_hardware_support/pilz_manipulator_mock.h
martiniil: It also mocks the canopen_chain_node.

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
```diff
@@ -15,94 +15,386 @@
...
+   * Access to the state machine is protected for thread-safety.
+   *
+   * @param sto The updated sto value.
+   */
+  void updateSto(const bool sto);
+
+  /**
+   * @brief Call the service triggering recover of the drives.
+   */
+  void call_recover();
```
agutenkunst: Please compare comment and function name. How about `callRecoverService`

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
```diff
@@ -15,94 +15,386 @@
...
+
+  /**
+   * @brief Call the service triggering halt of the drives.
+   */
+  void call_halt();
```
agutenkunst: See above

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
```diff
@@ -15,94 +15,386 @@
...
+
+  /**
+   * @brief Call the service triggering hold of the controller. Wait until execution of the hold trajectory finished.
+   */
+  void call_hold();
```
agutenkunst: See above

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
```diff
@@ -15,94 +15,386 @@
...
+
+  /**
+   * @brief Call the service triggering unhold of the controller.
+   */
+  void call_unhold();
```
agutenkunst: See above

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
```diff
@@ -15,94 +15,386 @@
...
-  void performStop();
 
   /**
-   * @brief Update stored STO value and react to STO changes.
-   * Reactions:
-   *  - STO == false:   Perform Stop 1
-   *  - STO == true:    unhold Controller + recover Drives
+   * @brief Stop the state machine.
+   *
+   * @note This is only used in tests.
```
agutenkunst: No this is also done in the destructor

prbt_hardware_support/include/prbt_hardware_support/sto_state_machine.h
```diff
@@ -0,0 +1,342 @@
...
+#include <boost/msm/front/state_machine_def.hpp>
+
+namespace prbt_hardware_support
+{
+
+/**
+ * @brief An AsyncStoTask is represented by a task execution and a completion signalling.
+ *
+ * The separation of task execution and the completion signalling allows the task execution to be done asynchronously.
+ * Both functions have the signature @code void() @endcode.
```
agutenkunst, line 54: I can see that ;-)

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
```diff
@@ -15,94 +15,386 @@
...
 
 private:
-  boost::optional<bool> sto_ {boost::none};
+  //! State machine
+  std::unique_ptr<StoStateMachine> state_machine_;
+
+  //! Flag indicating if the worker-thread should terminate
+  std::atomic_bool terminate_{false};
+
+  //! Mutex for protecting access to the state machine
```
agutenkunst: We should more carefully explain where this mutex **needs** to be used

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
```diff
@@ -15,94 +15,386 @@
...
+  {
+    std::lock_guard<std::mutex> lock(sm_mutex_);
+    ROS_DEBUG("trigger sto_updated");
+    state_machine_->process_event(typename StoStateMachine::sto_updated(sto));
+  }
+  worker_cv_.notify_one();
 }
-#endif // ADAPTER_STO_H
+template <class T>
+void AdapterStoTemplated<T>::workerThreadFun()
```
agutenkunst, line 223: Could be directly a lambda on construction

prbt_hardware_support/include/prbt_hardware_support/sto_state_machine.h
```diff
@@ -0,0 +1,342 @@
...
+   */
+  void execute()
+  {
+    operation_();
+  }
+
+  /**
+   * @brief Signal completion of the task execution.
+   */
+  void signalCompletion()
```
agutenkunst, line 75: "Signal"?

prbt_hardware_support/spec.dox
```diff
@@ -27,28 +27,35 @@
...
+ * @note A Stop 1 consists of holding the controller and halting the drives afterwards.
+ *  - Holding the controller starts with performing a stop motion (possibly replacing other prevailing motions).
+ *  - As long as the controller is holded, no new motions can be performed.
+ *
+ * @note The term "recovering" in the following section refers to a pending call on the respective service.
  *
  * @definereq{Stop1_Trigger}
  *  - A Stop 1 is triggered, if at least one of the following cases is true:
- *    - the state of the STO changes from TRUE -> FALSE,
+ *    - the state of the STO changes from TRUE -> FALSE when the drives are not recovering,
```
agutenkunst: Added link to relevant Spec, @martiniil please check

prbt_hardware_support/test/integrationtests/integrationtest_modbus_adapter_sto.cpp
```diff
@@ -208,50 +218,59 @@ TEST_F(ModbusAdapterStoTest, testSetupNoDisableService)
 {
   manipulator_.advertiseHoldService(nh_, HOLD_SERVICE_T);
   manipulator_.advertiseUnholdService(nh_, UNHOLD_SERVICE_T);
+  manipulator_.advertiseIsExecutingService(nh_, IS_EXECUTING_SERVICE_T);
   manipulator_.advertiseRecoverService(nh_, RECOVER_SERVICE_T);
 
-  auto t = asyncConstructor();
-  ros::Duration(5.5).sleep(); // slightly longer than WAIT_FOR_SERVICE_TIMEOUT_S
+  auto adapter_thread = asyncConstructor();
+  ros::Duration(1.0).sleep();
```
agutenkunst, line 220: Why suddenly less sleep?

prbt_hardware_support/test/integrationtests/integrationtest_modbus_adapter_sto.cpp
```diff
@@ -318,46 +340,125 @@ TEST_F(ModbusAdapterStoTest, testRemoveUnholdService)
...
+  std::thread pub_thread = std::thread([this, &keep_publishing](){
+    ros::Rate rate(STO_CLEAR_PUBLISHING_RATE);
+    while (keep_publishing)
+    {
+      this->pub_.publish(this->createDefaultStoModbusMsg(STO_CLEAR));
+      rate.sleep();
+    }
+    });
+
+  BARRIER("unhold_callback");
```
agutenkunst: If I understand this correctly we support unholding if we know there is no Hold Service?
This is also not implied by the Spec Item "Controller_service_hold_optional"

prbt_hardware_support/test/integrationtests/integrationtest_modbus_adapter_sto.cpp
```diff
@@ -318,46 +340,125 @@ TEST_F(ModbusAdapterStoTest, testRemoveUnholdService)
...
```
agutenkunst, line 378: Not specified by "Controller_service_is_executing_optional" and needs to be discussed! Does not seem like a good idea to me

prbt_hardware_support/test/integrationtests/integrationtest_modbus_adapter_sto.cpp
```diff
@@ -524,74 +674,53 @@ TEST_F(ModbusAdapterStoTest, testNoSto)
...
-
-  virtual ~TestingAdapterSto() = default;
-
-public:
-  void testingUpdateSto(const bool sto)
-  {
-    updateSto(sto);
-  }
-};
+  ModbusMsgStoWrapperException* exception = new ModbusMsgStoWrapperException("test");
```
agutenkunst: Please use shared pointer. Maybe seems silly a first but a codebase without a delete / new would be nice to have.

prbt_hardware_support/test/integrationtests/integrationtest_stop1.cpp
```diff
@@ -231,12 +234,13 @@ TEST_F(Stop1IntegrationTest, testServiceCallbacks)
...
    * Step 3 *
    **********/
-  BARRIER("recover_callback");
+  BARRIER("unhold_callback");
   {
     InSequence dummy;
 
     // Call from Disconnect
     EXPECT_CALL(manipulator, holdCb(_,_)).Times(1);
+    EXPECT_CALL(manipulator, isExecutingCb(_,_)).Times(AtLeast(1));
```
agutenkunst: Why don't we know the exact time this happens?

prbt_hardware_support/test/unittests/unittest_adapter_sto.cpp
```diff
@@ -0,0 +1,827 @@
...
+
+/**
+ * @brief Test enabling, stopping and enabling again
+ *
+ * Test Sequence:
+ *  1. Run the sto adapter and call updateSto(true)),
+ *     let recover and unhold services return success
+ *  2. Call updateSto(false)),
+ *     let hold and halt services return success, let is_executing service return "executing" once and then "not executing"
+ *  3. Call updateSto(true)) repeatedly,
```
agutenkunst, line 145: Please explain a bit more the reason behind the spaming...

prbt_hardware_support/test/unittests/unittest_adapter_sto.cpp
```diff
@@ -0,0 +1,827 @@
...
+  EXPECT_CALL(mock_factory_, call_named(HOLD_SERVICE, _)) \
+      .WillOnce(DoAll(ACTION_OPEN_BARRIER_VOID(HOLD_SRV_CALLED_EVENT), Return(true)))
+
+#define EXPECT_HALT                                       \
+  EXPECT_CALL(mock_factory_, call_named(HALT_SERVICE, _)) \
+      .WillOnce(DoAll(ACTION_OPEN_BARRIER_VOID(HALT_SRV_CALLED_EVENT), Return(true)))
+
+#define EXPECT_IS_EXECUTING                                  \
+  EXPECT_CALL(mock_factory_, call_named(IS_EXECUTING_SERVICE, _)) \
+      .WillOnce(Invoke(isExecutingInvokeAction(true))) \
```
agutenkunst: Why return true and then false? Better split into two EXPECT?

prbt_hardware_support/test/unittests/unittest_adapter_sto.cpp
agutenkunst: Seems to me critical to use parallel threads in the test. Would be much more deterministic to run all the tests in parallel.

prbt_hardware_support/test/unittests/unittest_adapter_sto.cpp
```diff
@@ -0,0 +1,827 @@
+/*
```
agutenkunst, line 1: Various TestCases are printing
```
GMOCK WARNING:
Uninteresting mock function call - returning default value.
    Function call: handle_invalid_named(@0x7fded4f9c760 "manipulator_joint_trajectory_controller/is_executing")
          Returns: false
NOTE: You can safely ignore the above warning unless this call should not happen.  Do not suppress it by blindly adding an EXPECT_CALL() if you don't mean to enforce the call.  See https://github.com/google/googletest/blob/master/googlemock/docs/CookBook.md#knowing-when-to-expect for details.
```
Please fix

agutenkunst: Doing `stress --cpu 8 --io 4 --vm 2` and `for i in {0..100}; do catkin_make run_tests_prbt_hardware_support_gtest_unittest_adapter_sto; done`

results sometimes in stuck testcase of either one of
- AdapterStoTest.testSpamEnablePlusStop
- AdapterStoTest.testIsExecutingDelayedFail
- AdapterStoTest.testIsExecutingFail


Sadly I cannot reproduce with Level DEBUG showing the state transition, but this indicates some timing issue where I would suspect the spaming to be involved...

agutenkunst (changes requested):
I fix some minor things myself, see commit... Please take care about the instable tests

prbt_hardware_support/test/unittests/unittest_adapter_sto.cpp
agutenkunst: I suggest dropping the spam. I figure you needed it to achieve full coverage?

agutenkunst: Simply changing the debug level (log4j.logger.ros.prbt_hardware_support.STOStateMachine=DEBUG) has a huge impact on the test result. This unittest should be **re-written in a purely deterministic way**.

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
martiniil: Done

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
martiniil: Done

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
martiniil: Done

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
martiniil: Done

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
martiniil: Yes, I reformulated it.

prbt_hardware_support/include/prbt_hardware_support/sto_state_machine.h
martiniil: Of course you can ;-)

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
martiniil: Done

prbt_hardware_support/spec.dox
martiniil: Thanks! Looks good.

prbt_hardware_support/include/prbt_hardware_support/sto_state_machine.h
martiniil: Why not?

prbt_hardware_support/include/prbt_hardware_support/adapter_sto.h
martiniil: I know. But I would prefer code readability over micro-optimization in this case.

prbt_hardware_support/test/integrationtests/integrationtest_modbus_adapter_sto.cpp
martiniil: Due to `WAIT_FOR_SERVICE_TIMEOUT_S=5.0` this test checked that there is no timeout. However the constant no longer exists. So the duration is arbitrarily chosen.

prbt_hardware_support/test/integrationtests/integrationtest_modbus_adapter_sto.cpp
martiniil: Done.

prbt_hardware_support/test/integrationtests/integrationtest_modbus_adapter_sto.cpp
martiniil: See above.

prbt_hardware_support/test/integrationtests/integrationtest_modbus_adapter_sto.cpp
martiniil: Good point. But there is also no contrary statement in the Spec.

Needs to be discussed. Probably the Spec has to be refined. I see a few options:
- check `hold_srv_client_.exists()` before calling unhold. But what if the preceding hold failed for another reason?
- deny all attempts to unhold once a controller service failed. 
- make all services required? That is always shut down when something unexpected happens?

What do you think?

prbt_hardware_support/test/integrationtests/integrationtest_stop1.cpp
martiniil: Well, because the service is called repeatedly. But I guess it was not designed for that and potentially leads to instabilities ...

After thinking it through I see the problem caused by the controller implementation. In my opinion the controller should either
- provide means for waiting until the hold position is reached
- wait until hold position is reached before returning the hold service. PR: #212

In order to make progress I will proactively use #212 to resolve the issues with is_executing in this context.

prbt_hardware_support/test/unittests/unittest_adapter_sto.cpp
martiniil: WIll be outdated if the controller change gets accepted.

prbt_hardware_support/test/unittests/unittest_adapter_sto.cpp
martiniil: The reason is that an sto update can happen anytime during a service call. But it should have no effect, if the sto didn't change.

Concerning your second comment: I do not see a need to add more parallel threads in the tests. So KISS

prbt_hardware_support/test/unittests/unittest_adapter_sto.cpp
martiniil: The problems seem to be resolved after removing the is_executing calls. I would not re-write the tests since I really focused on being deterministic.
However if you can figure a concrete part of the test that is not deterministic, please show me.

prbt_hardware_support/test/integrationtests/integrationtest_modbus_adapter_sto.cpp
martiniil: Or just print out a warning stating that a Stop1 might no longer be feasible? And that the hardware might be damaged by further stops?
After all we could never guarantee a Stop1. So we shouldn't make a big fuss about it, right?

prbt_hardware_support/test/unittests/unittest_adapter_sto.cpp
agutenkunst: Ok. First there was i missunderstanding. I am critical of having two test-threads and not just one (KISS)

prbt_hardware_support/test/unittests/unittest_adapter_sto.cpp
agutenkunst: Having tests fail here and there suggested some kind of non deterministic behaviour. I will have a closer look at it.

agutenkunst: At 34aa704b5e66fc2c25dd1c033cbeea8cc7f6ca93

I ran `for i in {0..100}; do catkin_make run_tests_prbt_hardware_support_gtest_unittest_adapter_sto; done`
an immediatly got stuck with:
- AdapterStoTest.testSpamEnablePlusStop

agutenkunst: Also

- AdapterStoTest.testUnholdFail get stuck

agutenkunst:
Just a short update before closing. Now 
```
[DEBUG] [1566306953.779437048]: updateSto(true)
[DEBUG] [1566306953.779504680]: Event: sto_updated - Leaving: RobotInactive
[DEBUG] [1566306953.779529426]: Event: sto_updated - Action: recover_start
[DEBUG] [1566306953.779558091]: Event: sto_updated - Entering: Enabling
[DEBUG] [1566306953.779640250]: Calling Recover (Service: /prbt/driver/recover)
[ INFO] [1566306953.781137748]: Recovering XXX
EMCY: 83#0000000000000000
[DEBUG] [1566306953.875349591]: updateSto(false)
[DEBUG] [1566306953.875440075]: Event: sto_updated - Leaving: Enabling
[DEBUG] [1566306953.875489740]: Event: sto_updated - Entering: StopRequestedDuringEnable
EMCY: 84#0000000000000000
EMCY: 85#0000000000000000
EMCY: 86#0000000000000000
EMCY: 87#0000000000000000
EMCY: 88#0000000000000000
[DEBUG] [1566306953.889808154]: Event: recover_done - Leaving: StopRequestedDuringEnable
[DEBUG] [1566306953.889886441]: Event: recover_done - Action: halt_start
[DEBUG] [1566306953.889918936]: Event: recover_done - Entering: Stopping
[DEBUG] [1566306953.889961051]: Calling Halt on driver (Service: /prbt/driver/halt)
[ INFO] [1566306953.891502354]: Halting down XXX
[DEBUG] [1566306953.891690906]: Event: halt_done - Leaving: Stopping
[DEBUG] [1566306953.891762494]: Event: halt_done - Entering: RobotInactive
```
which are **11ms delay** between recover and halt

agutenkunst (approved):

PR #173, opened 2019-07-08
sporadic test failure of prbt_hardware_support/rostest-test_unittests_unittest_filter_pipeline

ct2034:
@hslusarek 
https://docs.ros.org/kinetic/api/roscpp/html/classros_1_1NodeHandle.html#a81d08224944993c4cd38589e68417e12

PR #182, opened 2019-07-15
Fix incorrect usage of ros::Rate in test

martiniil:
... in order to stabilize unittest_pilz_modbus_client.test

see: https://github.com/PilzDE/pilz_robots/pull/173#issuecomment-510926318


agutenkunst (approved):
lgtm

PR #184, opened 2019-07-18
Make hold call of controller invariant

agutenkunst:
* The second call to hold should return immediatly as
  there is nothing todo. Helps solving #151

### DIFF
--- a/pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
+++ b/pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
@@ -25,9 +25,6 @@ namespace ph = std::placeholders;
 template <class SegmentImpl, class HardwareInterface>
 PilzJointTrajectoryController<SegmentImpl, HardwareInterface>::
 PilzJointTrajectoryController()
-  : active_update_strategy_(
-      std::bind(&PilzJointTrajectoryController::updateStrategyWhileHolding, this, ph::_1, ph::_2, ph::_3)
-    )
 {
 }
 
@@ -96,11 +93,18 @@ bool PilzJointTrajectoryController<SegmentImpl, HardwareInterface>::
 handleHoldRequest(std_srvs::TriggerRequest&, std_srvs::TriggerResponse& response)
 {
   std::lock_guard<std::mutex> lock(sync_mutex_);
+
+  if(active_mode_ == Mode::HOLD)
+  {
+    response.message = "Already in hold mode";
+    response.success = true;
+    return true;
+  }
+
+  active_mode_ = Mode::HOLD;
+
   JointTrajectoryController::preemptActiveGoal();
   triggerMovementToHoldPosition();
-
-  active_update_strategy_ = std::bind(&PilzJointTrajectoryController::updateStrategyWhileHolding, this,
-                                      ph::_1, ph::_2, ph::_3);
 
   response.message = "Holding mode enabled";
   response.success = true;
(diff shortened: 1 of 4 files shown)


pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
```diff
@@ -95,13 +95,23 @@ template <class SegmentImpl, class HardwareInterface>
...
 handleHoldRequest(std_srvs::TriggerRequest&, std_srvs::TriggerResponse& response)
 {
+  if(active_mode_ == Mode::HOLD)
+  {
+    response.message = "Already in hold mode";
+    response.success = true;
+    return true;
+  }
+
   std::lock_guard<std::mutex> lock(sync_mutex_);
```
jschleicher: Why don't you lock_guard this shortcut? If a parallel service call to `/unhold` is pending but hasn't set the `active_mode_` yet, you might get a race condition.

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
```diff
@@ -95,13 +95,23 @@ template <class SegmentImpl, class HardwareInterface>
...
+
   JointTrajectoryController::preemptActiveGoal();
   triggerMovementToHoldPosition();
 
   active_update_strategy_ = std::bind(&PilzJointTrajectoryController::updateStrategyWhileHolding, this,
```
jschleicher: Since you introduce a state now, you could replace this updateStrategy function pointer with a more readable `if` block inside a single update function. (Get rid of `active_update_strategy_` in favor of using `active_mode_`)

jschleicher (commented):
See inline comments.

martiniil (changes requested):
see Joachims comments. updateTrajectoryCommand can be refactored.

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
agutenkunst: Correct. Will fix

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
agutenkunst: Yep :+1:

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
```diff
@@ -133,7 +144,13 @@ bool PilzJointTrajectoryController<SegmentImpl, HardwareInterface>::
...
 {
   std::lock_guard<std::mutex> lock(sync_mutex_);
-  return active_update_strategy_(msg, gh, error_string);
+  if(active_mode_ == Mode::HOLD)
+  {
+    return updateStrategyWhileHolding(msg, gh, error_string);
+  }
+
+  // The default case
+  return updateStrategyDefault(msg, gh, error_string);
```
martiniil, line 153: This is just the call of the base class method. I would remove `updateStrategyDefault()`, too.

martiniil (dismissed):

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
jschleicher: `updateStrategyWhileHolding` and `updateStrategyDefault` are one-liners. I'd suggest to inline them here directly (making the warning and the base class function call explicit).

PR #186, opened 2019-07-19
drop unused lapack dependency

jschleicher:
The dependency is not needed by our solver so I propose to drop it.


ct2034 (changes requested):
Can you document this somewhere, so that we know, what manual steps to take whenever we generate the code again?

jschleicher:
@ct2034 Added a quilt series to auto-apply the patches after the auto-update using your script.
No additional manual steps are necessary any more.

prbt_ikfast_manipulator_plugin/update_ikfast_plugin.sh
```diff
@@ -7,7 +7,7 @@ planning_group_name=manipulator
 ikfast_plugin_pkg=prbt_ikfast_manipulator_plugin
 base_link_name=prbt_base_link
 eef_link_name=prbt_flange
-ikfast_output_path=/home/ch/ros/pilz_ws/src/pilz_robots/prbt_ikfast_manipulator_plugin/src/prbt_manipulator_ikfast_solver.cpp
```
ct2034, line 10: This is also written by the script .. https://github.com/ros-planning/moveit/blob/master/moveit_kinematics/ikfast_kinematics_plugin/scripts/create_ikfast_moveit_plugin.py#L316
Maybe we could change that....

ct2034 (approved):
Cool! :+1:

prbt_ikfast_manipulator_plugin/update_ikfast_plugin.sh
jschleicher: Indeed. Feel free to file a PR upstream.

jschleicher:
@ct2034 ping
Could this be merged now?

Issue #190, opened 2019-07-24
unittest_libmodbus_client can get stuck


agutenkunst:
### Commit
1b838d205685274311bcc3b6c18301841c41fae0

### Steps to reproduce
` for i in {0..100}; do GTEST_FILTER=*testNegativeNumberOfRegistersToRead* catkin_make run_tests_prbt_hardware_support_gtest_unittest_libmodbus_client; done`

will eventually get stuck

### Expected behavior
Test runs smoothly

### Observed behavior
Test gets stuck will eventually timeout -> No result for test -> Travis fail

### Reason
Seems to be some issue with closing the socket.

### Logfiles
Note: Set the LogLevel for **ros.prbt_hardware_support.ServerMock** and **ros.prbt_hardware_support.LibModbusClient** to debug in order to get the same output.

```
Note: Google Test filter = *testNegativeNumberOfRegistersToRead*
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from LibModbusClientTests
[ RUN      ] LibModbusClientTests.testNegativeNumberOfRegistersToRead
[DEBUG] [1563966280.797337585 [/home/agutenkunst/catkin_ws/src/pilz_robots/prbt_hardware_support/test/unittests/pilz_modbus_server_mock.cpp:65]: Starting Listening on Port 127.0.0.1:20500
[DEBUG] [1563966280.797391176 [/home/agutenkunst/catkin_ws/src/pilz_robots/prbt_hardware_support/test/unittests/pilz_modbus_server_mock.cpp:164]: About to start to listen for a modbus connection
[ INFO] [1563966280.797453646 [/home/agutenkunst/catkin_ws/src/pilz_robots/prbt_hardware_support/test/unittests/pilz_modbus_server_mock.cpp:174]: Waiting for connection...
[DEBUG] [1563966280.797489439 [/home/agutenkunst/catkin_ws/src/pilz_robots/prbt_hardware_support/test/unittests/pilz_modbus_server_mock.cpp:107]: Set Modbus data for Modbus-Server...
[DEBUG] [1563966280.797515788 [/home/agutenkunst/catkin_ws/src/pilz_robots/prbt_hardware_support/test/unittests/pilz_modbus_server_mock.cpp:114]: Modbus data for Modbus-Server set.
The client connection from 127.0.0.1 is accepted
[...]

PR #191, opened 2019-07-24
Extend logging of unittest_libmodbus_client

agutenkunst:
* Create named logging for ModbusServerMock and LibModbusClient
* Change the logger level in the test (tbd: find better way)
* Was needed to find https://github.com/PilzDE/pilz_robots/issues/190 (will open PR as soon as this is done...)

### DIFF
--- a/prbt_hardware_support/src/libmodbus_client.cpp
+++ b/prbt_hardware_support/src/libmodbus_client.cpp
@@ -39,7 +39,7 @@ bool LibModbusClient::init(const char* ip, unsigned int port)
 
   if (modbus_connect(modbus_connection_) == -1)
   {
-    ROS_ERROR_STREAM("Could not establish modbus connection." << modbus_strerror(errno));
+    ROS_ERROR_STREAM_NAMED("LibModbusClient", "Could not establish modbus connection." << modbus_strerror(errno));
     modbus_free(modbus_connection_);
     modbus_connection_ = nullptr;
     return false;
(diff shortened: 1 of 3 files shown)


prbt_hardware_support/test/unittests/unittest_libmodbus_client.cpp
```diff
@@ -282,5 +284,16 @@ TEST(LibModbusClientTests, setResponseTimeout)
 int main(int argc, char *argv[])
 {
   testing::InitGoogleTest(&argc, argv);
+
+  //Set logger level for server mock
+  if( ros::console::set_logger_level("ros.prbt_hardware_support.ServerMock", ros::console::levels::Debug) ) {
```
ct2034: please dont set logger levels programmatically

ct2034 (changes requested):

prbt_hardware_support/test/unittests/unittest_libmodbus_client.cpp
agutenkunst: Really would like a better solution but for now and for tests this should be fine. As long as there is now way using environment variables with rosconsole...

agutenkunst: Suggestion?

agutenkunst: Ok so in case someone wants to set the loggerlevel for Travis/Industrial_CI see
https://gist.github.com/agutenkunst/2f96826b96696784eefb985f73222a58

agutenkunst: Removed the part

ct2034: What you did

ct2034 (approved):

PR #194, opened 2019-07-25
Inc. port of unittest_libmodbus_client testcases

agutenkunst:
* Tests are unstable and still get stuck. Maybe this will help.
* Hopefully closes #193

PR #196, opened 2019-07-29
Don't read unnecessary registers

ct2034:
We currently read two types of not necessary registers:
* Those that we only want to write
* Those that lie in between the ones we want to read	

This causes problems when we have no right to read one or more of the unnecessary registers.

We fix this by:
* separate api definitions for read and write
* only read modbus registers that are explicitly configured (not in a single block)


prbt_hardware_support/include/prbt_hardware_support/modbus_api_spec.h
```diff
@@ -29,13 +29,16 @@
...
 {
+static const std::string READ_API_SPEC_PARAM_NAME {"read_api_spec/"};
 
 namespace modbus_api_spec
 {
   static const std::string STO {"STO"};
   static const std::string VERSION {"VERSION"};
   static const std::string BRAKETEST_REQUEST {"BRAKETEST_REQUEST"};
   static const std::string OPERATION_MODE {"OPERATION_MODE"};
+  static const std::string BRAKETEST_PERFORMED {"BRAKETEST_PERFORMED"};
```
agutenkunst: Not needed in this PR

prbt_hardware_support/src/libmodbus_client.cpp
```diff
@@ -64,6 +64,7 @@ unsigned long LibModbusClient::getResponseTimeoutInMs()
 
 RegCont LibModbusClient::readHoldingRegister(int addr, int nb)
 {
+  ROS_DEBUG("readHoldingRegister()");
```
agutenkunst: Remove ?

prbt_hardware_support/src/pilz_modbus_client.cpp
```diff
@@ -177,4 +194,28 @@ void PilzModbusClient::run()
   state_ = State::not_initialized;
 }
 
+void PilzModbusClient::split_into_blocks(std::vector<std::vector<unsigned short>> &out, const std::vector<unsigned short> &in){
+  unsigned short prev{0};
+  std::vector<unsigned short> current_block;
+  for (auto & reg : in){
+    if(reg <= prev){
+      throw PilzModbusClientException("List must be sorted.");
```
agutenkunst: Or just sort in case... more user-friendly

prbt_hardware_support/src/pilz_modbus_client.cpp
```diff
@@ -177,4 +194,28 @@ void PilzModbusClient::run()
   state_ = State::not_initialized;
 }
 
+void PilzModbusClient::split_into_blocks(std::vector<std::vector<unsigned short>> &out, const std::vector<unsigned short> &in){
```
agutenkunst: return result?

prbt_hardware_support/src/pilz_modbus_client.cpp
```diff
@@ -177,4 +194,28 @@ void PilzModbusClient::run()
...
+  for (auto & reg : in){
+    if(reg <= prev){
+      throw PilzModbusClientException("List must be sorted.");
+    }
+    else if(reg == prev + 1) {
+      current_block.push_back(reg);
+    }
+    else { // *it >= prev + 1
+      std::vector<unsigned short> to_out(current_block);
+      if(to_out.size() > 0)
```
agutenkunst: braces

prbt_hardware_support/src/pilz_modbus_client.cpp
```diff
@@ -177,4 +194,28 @@ void PilzModbusClient::run()
...
+        out.push_back(to_out);
+      current_block.clear();
+      current_block.push_back(reg);
+    }
+    prev = reg;
+  }
+  std::vector<unsigned short> to_out(current_block);
+  if(to_out.size() > 0)
```
agutenkunst: braces

agutenkunst (changes requested):

agutenkunst (approved):

agutenkunst:
So there seems to be an issue on travis with this commit:
Screenshot from 2019-08-02 07-47-53

I will have a look at it...

agutenkunst:
`rosrun industrial_ci run_travis` 
and 
`catkin build --no-deps --catkin-make-args run_tests -- --verbose prbt_hardware_support` 

is running perfectly fine on my machine(using http://packages.ros.org/testing/ubuntu).

agutenkunst:
Update: Sadly #194 did not fix the issue...

agutenkunst:
See https://travis-ci.org/PilzDE/pilz_robots/jobs/566875454

PR #199, opened 2019-07-31
also allow sto:='None' in upper-case

jschleicher:
by case-insensitive string-compare in the launch eval expression


ct2034 (approved):
nice UX :+1:

PR #200, opened 2019-08-01
Ready code for sanitizing

HiddenBug:
Make code ready for [sanitizing](https://github.com/google/sanitizers). 
In order to do that, I firstly fixed some bugs which are shown by the the clang compiler. Consequently, I added the CMake flags needed to active the different sanitizers. The sanitizer flags are all off by default. Only one sanitizer flag can be activate at any time. In order to use the different sanitizer, the compiler has to be switched to the clang compiler. See [here](https://stackoverflow.com/questions/7031126/switching-between-gcc-and-clang-llvm-using-cmake/12843988#12843988) how to do this. If you now run the different tests with the different sanitizer, you will see some errors. Try it out and have fun ;)

### DIFF
--- a/pilz_control/CMakeLists.txt
+++ b/pilz_control/CMakeLists.txt
@@ -24,6 +24,20 @@ catkin_package(
 
 add_definitions(-std=c++11)
 
+# To switch compiler see: https://stackoverflow.com/questions/7031126/switching-between-gcc-and-clang-llvm-using-cmake/12843988#12843988
+
+# Works with g++ and clang compiler (Do not forget to build with "-DCMAKE_BUILD_TYPE=DEBUG" flag)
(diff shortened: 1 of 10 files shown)


pilz_control/CMakeLists.txt
```diff
@@ -24,6 +24,20 @@ catkin_package(
 
 add_definitions(-std=c++11)
 
+# To switch compiler see: https://stackoverflow.com/questions/7031126/switching-between-gcc-and-clang-llvm-using-cmake/12843988#12843988
```
agutenkunst, line 27: @HiddenBug Could you auto detect the compiler and only set the respective flags if CLANG is enabled?

agutenkunst (changes requested):

agutenkunst:
Also please resolve the conflict, afterwards I would be open for merging

pilz_control/CMakeLists.txt
HiddenBug: See PR #281

HiddenBug:
Continued in PR #281.

PR #201, opened 2019-08-01
Fix/clang tidy errors

HiddenBug:


HiddenBug:
@agutenkunst Please consider merging this PR. It will remove the clang-tidy warnings which also appear on the travis build farm.

prbt_hardware_support/src/pilz_modbus_client.cpp
```diff
@@ -129,17 +129,17 @@ void PilzModbusClient::run()
       }
     }
 
-    std::vector<std::vector<unsigned short>> blocks = split_into_blocks(registers_to_read);
+    std::vector<std::vector<unsigned short>> blocks = splitIntoBlocks(registers_to_read);
 
     unsigned short index_of_first_register = *std::min_element(registers_to_read.begin(), registers_to_read.end());
     int num_registers = *std::max_element(registers_to_read.begin(), registers_to_read.end()) - index_of_first_register + 1;
```
martiniil: Here is also an implicit conversion: unsigned short -> int

martiniil: Please rename `registers_to_read` -> `registers_to_read_`

martiniil (changes requested):
Thx. Great, no warnings left.

I have very little comments. I hope you don't mind.

prbt_hardware_support/src/pilz_modbus_client.cpp
HiddenBug: **Renaming:** Done -> See 67114e6bc29a0074893d12fae2997248b9a7740a

**Implicit conversion:** Is this an warning/error which is shown on your PC? I do not get this warning/error. However, I see what you mean. The conversion to `int` is because of the equation. Even if I change the type of `num_registers` to `unsigned short,` the conversion of the elements of the add operation to `int` remains. If it is not a warning/error on your PC I suggest to leave it the way it is, because in that case it does not belong to this PR. What do you mean? Or maybe you have a solution?

hslusarek:
PR is continued in #218

PR #210, opened 2019-08-08
new package pilz_utils

martiniil:


leunMar:
please remove the WIP label

leunMar:
@agutenkunst Any idea why this is failing? When I ran the coverage build two times on my local machine, it succeeds the second time.

> Combining tracefiles.
> Reading tracefile pilz_utils_coverage.base
> lcov: ERROR: no valid records found in tracefile pilz_utils_coverage.base

leunMar:
Never mind ... I am after it.

pilz_utils/include/pilz_utils/wait_for_topic.h
```diff
@@ -0,0 +1,48 @@
...
+#include <pilz_utils/wait_for_timeouts.h>
+
+namespace pilz_utils
+{
+
+/**
+ * @brief Waits until someone publishes on the specified topic.
+ */
+template<class T>
+static void waitForTopic(const std::string topic_name,
```
agutenkunst, line 34: This function should really have a better name. See the diff in comment and function name?
However this can be done in a seperate refactoring.

prbt_hardware_support/src/brake_test_executor_node.cpp
```diff
@@ -34,6 +35,7 @@ static const std::string CONTROLLER_HOLD_MODE_SERVICE_NAME{"/prbt/manipulator_jo
 static const std::string CONTROLLER_UNHOLD_MODE_SERVICE_NAME{"/prbt/manipulator_joint_trajectory_controller/unhold"};
 static const std::string BRAKE_TEST_RESULT_SERVICE_NAME{"/prbt/send_brake_test_result"};
 
+using namespace pilz_utils;
```
agutenkunst, line 38: I would be in favor of addressing the function directly with namespace or at least defining an alias if you consider "pilz_utils" to be to long.
I hope that generally we did not "using" on other packages that often. I like the fact that this hints that another package is used because especially if this package lives in another repo some day this could be useful

agutenkunst (changes requested):
Just 2 things. Otherwise looks good. This will be a separate repo some day right?

agutenkunst:
Also please rebase.

prbt_hardware_support/src/brake_test_executor_node.cpp
leunMar: I changed it.

leunMar: 42e6b16d3497e82f8a976b0c5691ac5a3eb9cd89

pilz_utils/include/pilz_utils/wait_for_topic.h
leunMar: Done: https://github.com/PilzDE/pilz_robots/pull/278

PR #212, opened 2019-08-14
Move to hold position during hold service

martiniil:
I see no benefit in returning immediately. I also think that it was counter-intuitive.


ct2034:
@hslusarek What do you think?

hslusarek:
@ct2034 The code changes are already in melodic-devel. Only the changelog change is missing. However the changelog was already released. Therefore, I suggest to close the PR without merging. What do you think?

ct2034:
This is weird. Yes, ok. I close it.

PR #221, opened 2019-08-22
A wildcard to exclude all generated headers from coverage

ct2034:
We have had this list of headers to exclude. It is cleaner this way. 
The only thing is that this also excludes some files that we previously tested (which were green). Is this a good idea? @hslusarek @agutenkunst @martiniil


ct2034:
### before:
Screenshot from 2019-08-22 15-51-10

### after:
Screenshot from 2019-08-22 15-46-14

hslusarek:
> We have had this list of headers to exclude. It is cleaner this way.
> The only thing is that this also excludes some files that we previously tested (which were green). Is this a good idea?

In my opinion all auto generated files can be excluded. Are files excluded which are implemented by us? If so, which ones?

ct2034:
No. None were implemented by us. Can you please approve, then.

hslusarek (approved):

ct2034:
another case of #230

PR #229, opened 2019-08-29
Fixing the unittest filter_pipeline

ct2034:


ct2034:
one instability remains .. (https://travis-ci.org/PilzDE/pilz_robots/jobs/578411138#L5011)
```
<failure message="/root/catkin_ws/src/pilz_robots/prbt_hardware_support/test/unittests/unittest_pilz_modbus_client.cpp:518&#x0A;      Expected: expected_value&#x0A;      Which is: 3&#x0A;To be equal to: curr_value&#x0A;      Which is: 4&#x0A;Frequency used by PilzModbusClient is probably too high" type=""><![CDATA[/root/catkin_ws/src/pilz_robots/prbt_hardware_support/test/unittests/unittest_pilz_modbus_client.cpp:518
      Expected: expected_value
      Which is: 3
To be equal to: curr_value
      Which is: 4
Frequency used by PilzModbusClient is probably too high]]></failure>
    </testcase>
```

ct2034:
I think, we can conclude: It was my fault :grin:

PR #231, opened 2019-08-30
Introduce modbus brake test result adapter

hslusarek:


prbt_hardware_support/include/prbt_hardware_support/modbus_adapter_brake_test.h
```diff
@@ -19,28 +19,35 @@
...
 {
 
+using TRegVec = std::vector<uint16_t>;
+using TWriteModbusRegister = std::function<bool(const uint16_t&, const TRegVec&)>;
+
 /**
  * @brief Listens to the modbus_read topic and publishes a message
  * informing about a required brake test.
  */
-class ModbusAdapterBrakeTest : public AdapterBrakeTest
```
agutenkunst, line 37: Not a big fan of dropping the base class since it was probably introduced with something in mind

prbt_hardware_support/include/prbt_hardware_support/modbus_adapter_brake_test.h
```diff
@@ -19,28 +19,35 @@
...
 #include <prbt_hardware_support/ModbusMsgInStamped.h>
 #include <prbt_hardware_support/modbus_api_spec.h>
-#include <prbt_hardware_support/filter_pipeline.h>
+#include <prbt_hardware_support/IsBrakeTestRequired.h>
+#include <prbt_hardware_support/SendBrakeTestResult.h>
 
 namespace prbt_hardware_support
 {
 
+using TRegVec = std::vector<uint16_t>;
```
agutenkunst: I am in favor of dropping the "T" prefix since it does not provide useful information. For me this is something like a bit hungarian notation. https://en.wikipedia.org/wiki/Hungarian_notation

prbt_hardware_support/srv/SendBrakeTestResult.srv
```diff
@@ -0,0 +1,25 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Result of brake test to be sent
+uint16 performed
+uint16 result
```
agutenkunst: Was is the meaning of this value? Where is it defined?

agutenkunst (changes requested):
Please have a look at the comments.

agutenkunst:
Please have a look at the coverage.

prbt_hardware_support/srv/SendBrakeTestResult.srv
hslusarek: Simplified and documented SendBrakeTestResult service. -> Done

PR #235, opened 2019-09-04
Fix/sto state machine

### LINKED ISSUE #234: no_transition failure in StoStateMachine
### Commit
a44b3dfe34ea7bcfe7c524af9023e3accaf8cc87

### Steps to reproduce
Using the Stop1Executor, call updateSto(true) during hold service call and return

### Expected behavior
No failure.

### Observed behavior
Program execution is aborted.

### Log
```
/usr/include/boost/msm/front/state_machine_def.hpp:203: void boost::msm::front::state_machine_def<Derived, BaseState>::no_transition(const Event&, FSM&, int) [with FSM = boost::msm::back::state_machine<prbt_hardware_support::StoStateMachine_>; Event = prbt_hardware_support::StoStateMachine_::hold_done; Derived = prbt_hardware_support::StoStateMachine_; BaseState = boost::msm::front::default_base_state]: Assertion `false' failed.
```

martiniil:
Fix #234

- [x] check line and function coverage
- [x] rebase against #236 and check spec references in modified unittest


martiniil:
I get the following from the coverage report of sto_state_machine.h:

2019-09-04_screenshot

However the function is definitely called. I already tried adding `-fno-inline` to `COVERAGE_COMPILER_FLAGS` with no effect.

@agutenkunst any idea?

agutenkunst:
As discussed this missing coverage is caused by the Logging-Level

agutenkunst (approved):

PR #236, opened 2019-09-05
Fix sto spec

martiniil:
* a stop1 always is always completely executed
* stop1-release cannot be resumed (due to above requirement)

These specifications are already fullfilled by the current implementation.

### DIFF
--- a/prbt_hardware_support/spec.dox
+++ b/prbt_hardware_support/spec.dox
@@ -30,26 +30,11 @@
  * @note STO == 1/TRUE  <-> manipulator is allowed to move freely
  * @note STO == 0/FALSE <-> manipulator is NOT allowed to move (except for stop motion)
  *
- * @note A Stop 1 consists of holding the controller and halting the drives afterwards.
- *  - Holding the controller starts with performing a stop motion (possibly replacing other prevailing motions).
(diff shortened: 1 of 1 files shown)


prbt_hardware_support/spec.dox
```diff
@@ -54,19 +54,12 @@
...
- * system stops the stop1 process as soon as the currently active step is finished.
- * For example:
- *    If the state of the STO changes from TRUE -> FALSE while the drives
- *    are recovering, the controller is kept in "holding" mode and
- *    the execution of the stop trajectory is skipped.
- *
  * Please note:
  * - While the controller is holded, the controller rejects all new trajectories.
- * - The stop1 process can also be resumed by again chainging the STO to FALSE.
+ * - The stop1 process cannot be interrupted, even if the STO changes to TRUE.
```
agutenkunst, line 44: I think we should specify "stop1 process"...

agutenkunst (changes requested):
On thing...

prbt_hardware_support/spec.dox
martiniil: We would repeat ourselves because we already specified a "stop1 execution" ...
I will try to simplify everything

agutenkunst (approved):

PR #240, opened 2019-09-10
Fix generated headers exclude for coverage

martiniil:
As I understand it, `/devel/include/${PROJECT_NAME}` is where generated headers are usually located. The previous wildcard didn't work for me.


martiniil:
Reverting a44b3dfe34ea7bcfe7c524af9023e3accaf8cc87 is also fine.

PR #245, opened 2019-09-11
Remove services and topics from BrakeTestExecutor

hslusarek:


prbt_hardware_support/src/brake_test_executor.cpp
```diff
@@ -15,92 +15,85 @@
...
   }
-
-  BrakeTest braketest_srv;
-  if (!trigger_braketest_client_.call(braketest_srv))
+  else
   {
-    response.success = false;
-    response.error_msg = "Failed to trigger brake test via service " + trigger_braketest_client_.getService();
-    response.error_code.value = BrakeTestErrorCodes::TRIGGER_BRAKETEST_SERVICE_FAILURE;
+    ROS_WARN_STREAM("Calling hold on controller skipped because no hold service given");
```
agutenkunst: We need better error handling. If a hold is not possible the whole system is in a critical state...

agutenkunst: There is always true returned...

prbt_hardware_support/src/brake_test_executor_node.cpp
```diff
@@ -26,7 +44,28 @@ int main(int argc, char **argv)
   ros::init(argc, argv, "brake_test_executor");
   ros::NodeHandle nh{"~"};
 
-  prbt_hardware_support::BrakeTestExecutor brake_test_executor(nh);
+  waitForService(CONTROLLER_HOLD_MODE_SERVICE_NAME);
```
agutenkunst, line 47: What if the service is missing ?

prbt_hardware_support/src/brake_test_executor_node.cpp
```diff
@@ -26,7 +44,28 @@ int main(int argc, char **argv)
...
+  ros::ServiceClient hold_client = nh.serviceClient<std_srvs::Trigger>(CONTROLLER_HOLD_MODE_SERVICE_NAME);
+
+  waitForService(BRAKETEST_ADAPTER_SERVICE_NAME);
```
agutenkunst, line 50: What if the service is missing ?

prbt_hardware_support/src/brake_test_executor_node.cpp
```diff
@@ -26,7 +44,28 @@ int main(int argc, char **argv)
...
+  ros::ServiceClient brake_test_execute_client = nh.serviceClient<BrakeTest>(BRAKETEST_ADAPTER_SERVICE_NAME);
+
+  waitForService(CONTROLLER_UNHOLD_MODE_SERVICE_NAME);
```
agutenkunst, line 53: What if the service is missing ?

prbt_hardware_support/src/brake_test_executor_node.cpp
```diff
@@ -26,7 +44,28 @@ int main(int argc, char **argv)
...
+  ros::ServiceClient unhold_client = nh.serviceClient<std_srvs::Trigger>(CONTROLLER_UNHOLD_MODE_SERVICE_NAME);
+
+  waitForService(BRAKE_TEST_RESULT_SERVICE_NAME);
```
agutenkunst, line 56: What if the service is missing ?

agutenkunst (changes requested):
There are some questions but I can see that these were also relevant before this PR.

prbt_hardware_support/src/brake_test_executor.cpp
jschleicher: And this DOES change behaviour (hold was required before).

jschleicher: > There is always true returned...

Services always return true, if the result type was filled, see [ROS wiki](http://wiki.ros.org/roscpp/Overview/Services#Callback_Signature).

martiniil: Yes, I agree. Can we do the error handling separately?

prbt_hardware_support/src/brake_test_executor_node.cpp
martiniil: waits infinitely

prbt_hardware_support/src/brake_test_executor_node.cpp
martiniil: waits infinitely

prbt_hardware_support/src/brake_test_executor_node.cpp
martiniil: waits infinitely

prbt_hardware_support/src/brake_test_executor_node.cpp
martiniil: waits infinitely

prbt_hardware_support/src/brake_test_executor.cpp
martiniil: This PR should not change behaviour. I will correct it.

prbt_hardware_support/src/brake_test_executor_node.cpp
agutenkunst: I don't think so...

martiniil: waits until ros shutdown

PR #251, opened 2019-09-18
Feature/coverage ci

agutenkunst:
Needs full 100% coverage of prbt_hardware_support first.

Thus requires: #260 and #261

### DIFF
--- a/.coverage.sh
+++ b/.coverage.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+
+source "${ICI_SRC_PATH}/workspace.sh"
+source "${ICI_SRC_PATH}/util.sh"
+
+echo "Checking coverage for [$COVERAGE_PKGS]"
+coverage_pass=true
+
+for pkg in $COVERAGE_PKGS; do
+    echo "Creating coverage for [$pkg]"
+    ws=~/target_ws
+    extend="/opt/ros/$ROS_DISTRO"
+    ici_exec_in_workspace "$extend" "$ws" catkin build $pkg -v --no-deps --catkin-make-args ${pkg}_coverage
+    cd $TARGET_REPO_PATH
+    echo "Coverage summary for $pkg ----------------------"
+    lcov --summary $ws/build/$pkg/${pkg}_coverage.info.cleaned
+    echo "---------------------------------------------------"
+
+    line_cov_percentage=$(lcov --summary $ws/build/$pkg/${pkg}_coverage.info.cleaned 2>&1 | grep -Poi "lines\.*: \K[0-9.]*")
+    required_coverage="100.0"
+
+    if [ "$line_cov_percentage" != "$required_coverage" ]; then
+        result_str="$pkg: $line_cov_percentage%(required:$required_coverage%)\e[${ANSI_RED}m[failed]\e[0m"
+        echo -e $result_str
+        coverages+=("$result_str")
+        coverage_pass=false
+    else
+        result_str="$pkg($line_cov_percentage%)\e[${ANSI_GREEN}m[pass]\e[0m"
+        echo -e $result_str
+        coverages+=("$result_str")
+    fi
+
+done
+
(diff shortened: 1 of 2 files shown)


.coverage.sh
```diff
@@ -0,0 +1,41 @@
...
+    cd $TARGET_REPO_PATH
+    echo "Coverage summary for $pkg ----------------------"
+    lcov --summary $ws/build/$pkg/${pkg}_coverage.info.cleaned
+    echo "---------------------------------------------------"
+
+    line_cov_percentage=$(lcov --summary $ws/build/$pkg/${pkg}_coverage.info.cleaned 2>&1 | grep -Poi "lines\.*: \K[0-9.]*")
+    required_coverage="100.0"
+
+    if [ "$line_cov_percentage" != "$required_coverage" ]; then
+        echo "$pkg has $line_cov_percentage% line coverage";
```
ct2034: Please add a line, that says, the job fails because of this.. Maybe in red ...

.coverage.sh
```diff
@@ -0,0 +1,41 @@
...
+        coverages+=("$pkg: $line_cov_percentage% (required: $required_coverage%) \e[${ANSI_RED}m[failed]\e[0m")
+        coverage_pass=false
+    else
+        echo $pkg " has 100.0% line coverage";
```
ct2034: `        echo "$pkg has 100.0% line coverage";`
works too, right?
but its cleaner

.coverage.sh
```diff
@@ -0,0 +1,41 @@
...
+if [ "${coverages// }" != "" ]; then
+    echo "Coverage results:"
+    for coverages in "${coverages[@]}"; do
+        echo -e '  ' $coverages
+    done
+fi
+
+if ! $coverage_pass; then
+  exit 1;
+fi
```
ct2034: newline at the end of file missing

.coverage.sh
```diff
@@ -0,0 +1,41 @@
...
+    else
+        echo $pkg " has 100.0% line coverage";
+        coverages+=("$pkg ($line_cov_percentage%) \e[${ANSI_GREEN}m[pass]\e[0m")
+    fi
+
+done
+
+if [ "${coverages// }" != "" ]; then
+    echo "Coverage results:"
+    for coverages in "${coverages[@]}"; do
```
ct2034: why is this the same name?
coverages == coverages

.travis.yml
```diff
@@ -21,6 +21,11 @@ env:
   matrix:
     - ROS_REPO=ros
     - ROS_REPO=testing
+    - ROS_REPO="testing" \
+      CATKIN_LINT=false NOT_TEST_BUILD=true \
+      AFTER_SCRIPT="./.coverage.sh" \
+      CMAKE_ARGS="-DENABLE_COVERAGE_TESTING=ON -DCMAKE_BUILD_TYPE=Debug" \
+      DOCKER_RUN_OPTS='-e COVERAGE_PKGS="prbt_hardware_support"'
```
ct2034: is there a cleaner way?

ct2034 (changes requested):

.coverage.sh
agutenkunst: done

.coverage.sh
agutenkunst: Solved differently now. Please check again

.coverage.sh
agutenkunst: Line added. Correct?

.coverage.sh
agutenkunst: Fixed.

.travis.yml
agutenkunst: I am open for suggestions.

ct2034: I just thought it may be nicer to somehow wrap all of them into a file but that may increase the complexity unnecessarily and may not even be possible. So I guess we leave it as it is.

ct2034 (approved):

PR #254, opened 2019-09-26
WIP: Event-driven speed observer

martiniil:


martiniil:
tf time stamp of speed limit violation <-> observer detects speed limit violation < 10ms
observer detects speed limit violation <-> current trajectory is aborted ~15ms
current trajectory is aborted <-> tf time stamp of max speed ~80ms

Bildschirmfoto von 2019-09-30 10-06-50

martiniil:
Stop1, speed limit 0.2

rqt_plot_stop1

martiniil:
"Stop0", speed limit 0.2

rqt_plot_stop0

martiniil:
@hslusarek @ct2034 Have a look at the plots above. The frame speed and the actual joint velocity nicely align. However the actual state lags behind, so when we detect the limit violation, the controller is already way ahead.

If we could somehow compute the frame speeds from the desired joint states, we probably could stop in time and also avoid stopping trajectories that stay just below the limit.

martiniil:
Here the frame speeds where computed from the desired joint states of the controller.

The limit is 0.2, a limit violation is detected at ~0.22 and a stop1 is triggered at ~0.23. In the following execution of the hold trajectory the frame speed has a peek at ~0.26.

rqt_plot_stop1_desired_js

martiniil:
The time interval between the message callbacks is very fluctuating in the 10ms scale. This plot shows how fast the manipulator accelerates in comparison.

rqt_plot_accelerate

leunMar:
Closing in favor of #272

PR #269, opened 2019-10-16
Controller velocity check: Add first implementation

agutenkunst:
Still missing:

- [ ] List which checks are missing
- [ ] Check how this interacts with trajectory replacement
- [ ] Check with rqt_joint_trajectory_controller


leunMar:
If time_from_start is not zero, the current state is needed for computing the velocity...
However, accessing the current state inside the controller is not thread-safe.

leunMar:
It could be done by sampling at the current time. I am not really happy with that, because the sampling is later repeated in the original `updateTrajectoryCommand()`

leunMar:
Well, it is difficult to know the exact trajectory, the controller will create. It somehow creates a bridge-segment in `initJointTrajectory()` in order to attach to the current trajectory. I tried to approximate that, however I am not sure, if this is the right approach.

Maybe, if we could say that what we do is a worst-case approximation.

leunMar:
@agutenkunst I think we already talked about this, so I just wonder what is your opinion.

I came to the conclusion that the usage of the member `joints_` in the `joint_trajectory_controller` is not safe. If there is a *writing*-thread, there should not be anymore threads accessing (read/write) the container. Otherwise the behaviour is undefined. So a call to `joints_.size()` in `queryStateService()` or `setHoldPosition()` is not safe, because `update()` is writing here
```
  // Hardware interface adapter: Generate and send commands
  hw_iface_adapter_.updateCommand(time_data.uptime, time_data.period,
                                  desired_state_, state_error_);
```

edit: An argument is that changing the position in a JointHandle does not change `std::vector<JointHandle> joints_`?

leunMar:
Somehow the speeds computed by the controller and the tf observer differ from each other.
Maybe it is due to noise in the tf data?
https://answers.ros.org/question/12894/calculating-velocity-from-tf/?answer=19052#post-id-19052

leunMar:
Wow! tf noise:

speed_observer02

I think this *joint_states_speed_observer* I just committed is what we need as a tool to prove that our observer ansatz works.

leunMar:
I ran two tests. In the first one I command a PTP moving joint2. The following shows the frame speeds computed by the controller and the *acutal* frame speeds obtained through the joint_states from gazebo.
The max speed matches pretty good.

speed_observer03

leunMar:
In the second test I command joint2 with one click in rqt_joint_trajectory_controller. Again the frame speeds computed by the controller and the *actual* frame speeds (gazebo):

Well, now it would be good to know, if those oscillations also occur with the real robot ...

If I imagine the motion smoothed out, the *controller-computed* speeds are slightly lower. Probably this can be optimized (in case we decide for this ansatz) by choosing the starting point more intelligently.

speed_observer04

leunMar:
https://github.com/PilzDE/pilz_robots/tree/feature/controller_speed_check_on_update
limit: 0.2, full speed PTP:
speed_observer05

martiniil:
Investigation completed.

PR #271, opened 2019-10-17
Use pilz controller as standard for prbt in gazebo

martiniil:


jschleicher:
Please provide a rationale.

In my opinion gazebo has no need to use a hold service and the original controller is enough. At least as long as you do not have a PSS4000 simulator ;-)
Calling `/unhold` manually is tedious, if you just want to simulate robot dynamics.

leunMar:
This facilitates testing new features in gazebo.

There could be more features in the future (e.g. speed monitoring), which could be useful.

But I guess we could also use separate launch files.

jschleicher:
Yes, please use separate launch files in the `test/` directory therefore or add an argument with the current default.

PR #272, opened 2019-10-17
WIP: investigate speed observer

martiniil:
These changes simplify further investigation of the speed observation.
Start robot:
```
roslaunch prbt_gazebo sim_prbt_gazebo.launch pipeline:=pilz_command_planner
```
Start observer:
```
roslaunch prbt_hardware_support_speed_observer_node.launch sim:=true
```
Unhold controller:
```
rosservice call /prbt/manipulator_joint_trajectory_controller/unhold
```
Now you can start moving the robot and inspect e.g. with `rostopic echo /prbt/frame_speeds`.


leunMar:
actually you start without `sim:=true`

leunMar:
The delay we observed with the real robot is not present in gazebo simulation: (limit=0.2)

speed_observer01

leunMar:
But this can still be used to observe the actual frame speeds with a different observer ansatz.

leunMar:
Closing in favor of 4ba36e95a5601d9192af16d0571a4dc425a57c0c

PR #274, opened 2019-10-21
Set proper frequency in gazebo controller

agutenkunst:
Setting the controller frequency in gazebo to 8ms which should be a better match to the real system 
makes the robot oscillate. Could this be the reason why https://github.com/PilzDE/pilz_industrial_motion/issues/213 is working fine in simulation?

Does not seem to be the first time this is a problem. See https://github.com/ros-simulation/gazebo_ros_pkgs/issues/151. On startup gazebo even issues a warning.

@jschleicher your opinion?


leunMar:
This results in huge oscillations.
I did a small PTP motion and it already starts shaking severely.

jschleicher:
@agutenkunst  PilzDE/pilz_industrial_motion#213 is about a custom built real robot, so I don't see an interrelation. Why would you want to change the default gazebo control frequency in the first place?

agutenkunst:
> Why would you want to change the default gazebo control frequency in the first place?

Because reducing the diff between reality and simulation should help detect more errors at an early stage. Especially timing related behavior. But it seems that this is not done by simply setting the frequency to the real one.

PR #281, opened 2019-11-11
Add sanitizing build job

HiddenBug:
This PR adds sanitizing as CI jobs to the pilz_robots repository.


.travis.yml
```diff
@@ -26,6 +26,15 @@ env:
       AFTER_SCRIPT="./.coverage.sh"
       CMAKE_ARGS="-DENABLE_COVERAGE_TESTING=ON -DCMAKE_BUILD_TYPE=Debug"
       DOCKER_RUN_OPTS='-e COVERAGE_PKGS="pilz_control pilz_utils prbt_hardware_support"'
+    - ROS_REPO="testing"
+      CATKIN_LINT=false ADDITIONAL_DEBS="clang" CXX=clang++ CC=clang CMAKE_ARGS="-DCMAKE_BUILD_TYPE=Debug"
```
agutenkunst:
```suggestion
      CATKIN_LINT=false ADDITIONAL_DEBS="clang" CXX=clang++ CC=clang CMAKE_ARGS="-DCMAKE_BUILD_TYPE=Debug -DENABLE_ADDRESS_SANITIZING"
```

agutenkunst: Something like this could simplify the CMAKE checks

prbt_hardware_support/test/unittests/unittest_brake_test_executor.cpp
```diff
@@ -47,6 +49,31 @@ class SystemMock
 
 };
 
+
+
+TEST(BrakeTestExecutorTest, testAddressSanitizing)
```
agutenkunst: Just for clarification: This will be dropped finally?

prbt_hardware_support/test/unittests/unittest_speed_observer.cpp
```diff
@@ -148,14 +148,14 @@ MATCHER_P2(SpeedAtIGe, i, x,
            "Speed at index " + PrintToString(i) + std::string(negation ? "is not" : "is") + " greater or equal to" +
            PrintToString(x) + ".")
 {
-  return static_cast<unsigned long>(i) < arg->name.size() && arg->speed.at(i) >= x;
+  return static_cast<unsigned long>(i) < arg->name.size() && arg->speed.at(static_cast<unsigned long>(i)) >= x;
```
agutenkunst:
```suggestion
  return static_cast<unsigned long>(i) < arg->name.size() && arg->speed.at(static_cast<size_t>(i)) >= x;
```
maybe?

prbt_hardware_support/test/unittests/unittest_speed_observer.cpp
```diff
@@ -148,14 +148,14 @@ MATCHER_P2(SpeedAtIGe, i, x,
...
 }
 
 MATCHER_P2(SpeedAtILe, i, x,
            "Speed at index " + PrintToString(i) + std::string(negation ? "is not" : "is") + " less or equal to" +
            PrintToString(x) + ".")
 {
-  return static_cast<unsigned long>(i) < arg->name.size() && arg->speed.at(i) <= x;
+  return static_cast<unsigned long>(i) < arg->name.size() && arg->speed.at(static_cast<unsigned long>(i)) <= x;
```
agutenkunst: Same as above

agutenkunst (changes requested):
Really like having these checks included in the CI. Maybe even interesting for a upstream PR on industrial_ci. Just some minor remarks.

Just as an idea. This would be nice to have as a dedicated package just like https://github.com/mikeferguson/code_coverage for providing a simple macro to enable the sanitizing. However this is far beyond the scope of this PR. Maybe https://github.com/agutenkunst/cmake_extras_demo_pkg could be a starting point.

prbt_hardware_support/CMakeLists.txt
```diff
@@ -83,6 +83,21 @@ if(CATKIN_ENABLE_CLANG_TIDY)
 endif()
 
 
+################
+## Sanitizing ##
+################
+
+if(DEFINED ENV{ENABLE_ADDRESS_SANITIZING})
+  set (CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -g -fno-omit-frame-pointer -fsanitize=address")
```
agutenkunst:
```suggestion
  list (append CMAKE_CXX_FLAGS_DEBUG -g -fno-omit-frame-pointer -fsanitize=address")
```
should work as well. (Hold for the other "set" as well)

prbt_hardware_support/test/unittests/unittest_brake_test_executor.cpp
HiddenBug: Of course. Just for testing.

prbt_hardware_support/test/unittests/unittest_speed_observer.cpp
HiddenBug: Here we can see why macros s***. Normally you would use the container datatype itself to request the correct type for the index. Here it is not possible because it is a macro.

HiddenBug: Done.

prbt_hardware_support/test/unittests/unittest_speed_observer.cpp
HiddenBug: Done.

prbt_hardware_support/CMakeLists.txt
HiddenBug: Sorry does not work. I tried it but I get some strange errors which indicate that the list is not used in the same way as the set in that case. -> Done.

.travis.yml
HiddenBug: I agree. But what should I do now? Let's make these kind of improvements later on. One step after another ;) -> Done.

prbt_hardware_support/CMakeLists.txt
agutenkunst: So CMAKE_CXX_FLAGS_DEBUG does not seem to be a list but rather a string. Yep done.

.travis.yml
agutenkunst: Agree if you open a reminder issue

HiddenBug: See #282.

HiddenBug:
We could split the commits which are only needed to satisfy the clang compiler into a separate PR. @agutenkunst What do you think? See PR #283.

HiddenBug:
For information on how-to execute a sanitizer locally see PR pilzde/.github#7.

HiddenBug:
As it turns out gcc already supports address and undefined behavior sanitizing (see [gcc online documentation](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html)).
Furthermore, using clang introduces false-positive error detection's. This seems to happen when a library is not build with clang or something.
There also exists a thread sanitizer but I got a lot of errors using this sanitizer. So I decided to introduce the thread sanitizer later on and just introduce the address and undefined behavior sanitizer in this PR.

Memory sanitizing is only available with clang. However, _every_ dependency needs to be build with the memory sanitizing flags (described for example in the [rosproject-scripts repo on GitHub](https://github.com/hannob/rosproject-scripts/blob/63305fde06c9ddb31db1078686ec2f627253abbc/report/rosreport.tex#L46)). Therefore, I decided to skip this sanitizer completely (at least for the moment).

leunMar (commented):
:+1: for adding sanitizing

Why not use cmake arguments instead of environment variables for ENABLE_ADDRESS_SANITIZING, etc.?

HiddenBug:
> +1 for adding sanitizing
> 
> Why not use cmake arguments instead of environment variables for ENABLE_ADDRESS_SANITIZING, etc.?

Good question :smile:  I will give it a try.

HiddenBug:
> +1 for adding sanitizing
> 
> Why not use cmake arguments instead of environment variables for ENABLE_ADDRESS_SANITIZING, etc.?

OK. I only use the travis config file now. This should satisfy you.

HiddenBug:
> > +1 for adding sanitizing
> > Why not use cmake arguments instead of environment variables for ENABLE_ADDRESS_SANITIZING, etc.?
> 
> OK. I only use the travis config file now. This should satisfy you.

But now I get sanitizing errors in the repositories previously not included :see_no_evil: ARRRRRRRGH... I want to merge.

HiddenBug:
**Interesting fact:**
The address sanitizer detects a memory leak in `libclass_loader`, more precisely in `/opt/ros/melodic/include/class_loader/class_loader_core.hpp:205` (see [Travis log](https://api.travis-ci.org/v3/job/614764921/log.txt)). However, for some reason, maybe because the memory leak detection happens in another process as the test is running in, it does not lead to a test failure. I only see the error output message from the address sanitizer (or more precisely from the LeakSaniatizer).

martiniil:
Only guessing: from `class_loader_core.cpp`:
```
       // Insert into graveyard
        // We remove the metaobject from its factory map, but we don't destroy it...instead it
        // saved to a "graveyard" to the side.
        // This is due to our static global variable initialization problem that causes factories
        // to not be registered when a library is closed and then reopened.
        // This is because it's truly not closed due to the use of global symbol binding i.e.
        // calling dlopen with RTLD_GLOBAL instead of RTLD_LOCAL.
        // We require using the former as the which is required to support RTTI
```

This "graveyard" is a std::vector holding raw pointers. So maybe that is why these metaobjects are not destroyed.

martiniil:
I opened an issue: https://github.com/ros/class_loader/issues/131

PR #287, opened 2019-11-26
Fix setup for sim:=true. Fixes PilzDE/pilz_industrial_motion#230

martiniil:
@jschleicher correct?


jschleicher (changes requested):
It's unclear, why you put the two different cases into different hierarchies:

- `sim=True` stays in prbt_moveit_config
- `sim=False and has_operation_mode_support=False` is one level below in prbt_support

If you want to put the low-level launch stuff in low-level packages, I'd suggest creating a new launch file  `prbt_hardware_support/launch/simulate_safety.launch` and include it in `moveit_planning_execution.launch`.
Otherwise both `fake_speed_override_nodes` should stay in the same launch `moveit_planning_execution.launch` to avoid further confusion.

jschleicher (approved):
Thank you!

PR #288, opened 2019-11-26
WIP: get rid of redundant second braketest service

martiniil:
For some reason we assumed that all components are separate ROS nodes. This led to redundant communication and also ambiguous topics/service names.

I propose that we move away from this in manageable increments (like this one, hopefully).


agutenkunst (commented):
I like this idea and the visualization as a first increment. How about providing a simple class diagram for discussing the class relations. Imho having both adapter as template in the executor would provide great benefit during testing.
Would be up for a second increment directly after this is done.

hslusarek:
One advantage of the old concept, to use a separate node for the adapters, is the fact that you can "swap out" the real hardware via launch file to use simulated/mocked adapter simulating real hardware.

I can also use the same logic and ask: Why should we not allow the PythonAPI to include the `modbus_adapter_braketest_required` nodes code? Where do we stop with this logic? Maybe a simple renaming of the services is sufficient?

>  Imho having both adapter as template in the executor would provide great benefit during testing.

@agutenkunst Please explain this in more detail. If the executor is written with a good interface keeping the ROS-Framework at distance, I don't see how the inclusion of more code does improve the testing.

I suggest that we come up with a good and consistent overall concept explaining on how we do things and _why_  we do them this or that way, before "simply" changing things. @martiniil @agutenkunst What do you think?

PR #295, opened 2020-01-08
Test default value of fake speed override

### LINKED ISSUE #293: Incomplete system specification for speed monitoring/override
There are things missing in the system specification for the speed monitoring and the speed override in `prbt_hardware_support/spec.dox`. Probably also tests are missing.

- speed monitoring in case `has_operation_mode_support:=false` or `sim:=true`: robot is allowed to move with max speed.

- speed override in case `has_operation_mode_support:=false` or `sim:=true`: 1.0

martiniil:
Fixes #293.

### DIFF
--- a/prbt_hardware_support/spec.dox
+++ b/prbt_hardware_support/spec.dox
@@ -154,4 +154,10 @@
  *    - In T1 the speed override is 0.1.
  *    - In AUTOMATIC the speed override is 1.0.
  *    - If the operation mode is unknown, the speed override is set to zero.
+ *
+ *
+ * @subsection simulated_robot Simulated Robot
+ *
+ * @definereq{speed_override_for_simulated_robot}
+ * - If the system is connected to a simulated robot instead of a real robot the speed override (\ref speed_override) is 1.0.
  */
--- a/prbt_hardware_support/spec.dox
+++ b/prbt_hardware_support/spec.dox
@@ -156,8 +156,16 @@
  *    - If the operation mode is unknown, the speed override is set to zero.
  *
  *
- * @subsection simulated_robot Simulated Robot
+ * @subsection no_safety Other Setups (No ISO-10218-1 Support)
+ *
+ * @subsubsection simulated_robot Simulated Robot
  *
  * @definereq{speed_override_for_simulated_robot}
- * - If the system is connected to a simulated robot instead of a real robot the speed override (\ref speed_override) is 1.0.
+ * - If the system is connected to a simulated robot instead of a real robot, the speed override (\ref speed_override) is 1.0.
+ *
+ *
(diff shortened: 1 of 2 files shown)


prbt_hardware_support/spec.dox
```diff
@@ -154,4 +154,18 @@
...
  *    - In AUTOMATIC the speed override is 1.0.
  *    - If the operation mode is unknown, the speed override is set to zero.
+ *
+ *
+ * @subsection no_safety Other Setups (No ISO-10218-1 Support)
+ *
+ * @subsubsection simulated_robot Simulated Robot
+ *
+ * @definereq{speed_override_for_simulated_robot}
+ * - If the system is connected to a simulated robot instead of a real robot, the speed override (\ref speed_override) is 1.0.
```
HiddenBug, line 164: For now the formulation is ok because it simply reflects the state of our system. However, I strongly suggest that we introduce a mechanism to simulate the operation mode in simulation mode. Maybe we should open an issue for this?

prbt_hardware_support/spec.dox
```diff
@@ -154,4 +154,18 @@
...
+ *
+ *
+ * @subsubsection no_operation_mode No Operation Mode Support
+ *
+ * @definereq{speed_override_without_operation_mode_support}
+ * - If the system does not support operation modes, the speed override (\ref speed_override) is 1.0.
```
HiddenBug, line 170: For now the formulation is ok because it simply reflects the state of our system. 
However, based on our latest decisions regarding the supported FS controller variants, I suggest to completely drop the operation mode support option. Because all our FS controller have to support operation modes. If they currently do not support operation modes, then it is simply a hack. Ultimately,  my statement suggests to drop the support for all FS controller which currently do not support operation modes. If we can simulate the operation modes, then we can drop the operation mode support option completely because even the simulation mode can then by default support the operation mode.

prbt_hardware_support/spec.dox
HiddenBug: We should also discuss which parameter options (brake_test support, operation mode support, etc.) we want to support in simulation mode. -> Meeting?
I suggest to support only two variants:
- ISO `on`
- ISO `off`

HiddenBug (commented):
Nothing major. Just have a look at my comments. What do you think?

prbt_hardware_support/spec.dox
martiniil: Thanks for having a look. The meeting already took place. We decided against "simulated ISO".

prbt_hardware_support/spec.dox
martiniil: Thanks. We can discuss about that occasionally.

martiniil:
@HiddenBug From your comments I deduce that you approve the changes.

PR #298, opened 2020-01-13
Add TODO for supported hardware section

agutenkunst:


martiniil:
Was added through #324

Issue #187, opened 2019-07-22
Fix unstable unittest_pilz_joint_trajectory_controller


agutenkunst:
### Commit
https://github.com/PilzDE/pilz_robots/commit/69001be9805098286fa23096b3c9d05783b20b67

This test seems to fail on travis (especially the testIsExecuting) since it heavily relies on ros::Time::now()

martiniil:
You don't have a log by chance?
This could be related to the instability in #206, which seems to vanish or at least was heavily reduced through fd509a8d90da7f2bba1d43aa034da0267f9a788d.

agutenkunst:
Did not occurred again so far. Will close for now. Please reopen if needed.

agutenkunst:
Reopen ;-)
I just realized you could use sim time in this test.
See http://docs.ros.org/melodic/api/rostime/html/classros_1_1Time.html#ab569bffab97380d8a73275cdc02afb8f this would maybe even enable turning this into a unittest... 

@martiniil what do you think?

leunMar:
I would like this very much. But I don't agree on re-opening this issue. Or did you observe that the test is still unstable? If not, I would suggest that you either change the title of this issue (and remove the bug-label) or open a new issue.

leunMar:
Test failed again: (exceeds max time 60.0s) https://api.travis-ci.org/v3/job/601221025/log.txt

martiniil:
Should be fixed via fcc9c2c.

PR #302, opened 2020-01-20
Simulate time in unittest joint traj ctrl

HiddenBug:
In the past `unittest_pilz_joint_trajectory_controller` has provided unstable results (See #187). Deterministic test procedure can be achieved by using simulated time.
- [x] Refactor test using simulated time (`ros::Time::setNow`)
- [x] Introduce extended sleep function, which lets simulated time progress. Increases testability.


pilz_utils/include/pilz_utils/sleep.h
```diff
@@ -0,0 +1,48 @@
...
+ * \brief Extends ros::Duration::sleep() by increasing the simulation time, if needed.
+ * \note The usage of ros::Time in this function is not thread-safe.
+ */
+static bool sleep(const ros::Duration& sleep_duration)
+{
+  if (ros::Time::isSystemTime())
+  {
+    return sleep_duration.sleep();
+  }
+  else if (ros::Time::isValid())
```
HiddenBug: Just a minor thing: If you use `return` then `else` is basically redundant and can be removed.

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
```diff
@@ -106,7 +108,7 @@ handleHoldRequest(std_srvs::TriggerRequest&, std_srvs::TriggerResponse& response
   JointTrajectoryController::preemptActiveGoal();
   triggerMovementToHoldPosition();
 
-  ros::Duration(JointTrajectoryController::stop_trajectory_duration_).sleep();
+  pilz_utils::sleep(ros::Duration(JointTrajectoryController::stop_trajectory_duration_));
```
HiddenBug, line 111: You certainly have a reason why you change this line. So far, however, I'm missing the test which makes use of this change.

pilz_utils/include/pilz_utils/sleep.h
martiniil: Thanks.Done.

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
martiniil: I will push the changes on the test once they are completed.

martiniil: @HiddenBug ready for review

pilz_utils/include/pilz_utils/sleep.h
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (c) 2019 Pilz GmbH & Co. KG
```
HiddenBug: 2020?

pilz_utils/include/pilz_utils/sleep.h
```diff
@@ -0,0 +1,49 @@
...
+#define PILZ_UTILS_SLEEP
+
+#include <ros/ros.h>
+#include <ros/time.h>
+
+namespace pilz_utils
+{
+
+/**
+ * \brief Extends ros::Duration::sleep() by increasing the simulation time, if needed.
```
HiddenBug: For my personal taste, your description is a little bit short. Maybe just a few more words will help the reader understand, your intentions better. Something like:
"This functions does what ros::Duration:sleep() does but additionally increases the simulated time, in case a simulated clock is used."
Maybe you could also add a more extensive description explaining why this addition to the `ros::sleep` function is needed. What is it good for?
Also the description of the return type is missing. When does the function return `true` and when does it return `false`?

pilz_utils/include/pilz_utils/sleep.h
```diff
@@ -0,0 +1,49 @@
...
+  if (ros::Time::isSystemTime())
+  {
+    return sleep_duration.sleep();
+  }
+
+  if (ros::Time::isValid())
+  {
+    ros::Time::setNow(ros::Time::now() + sleep_duration);
+    return true;
+  }
```
HiddenBug, line 49: Is it possible to open a PR on `ros::Duration::sleep()` ([roscpp_core: sleep()](https://github.com/ros/roscpp_core/blob/kinetic-devel/rostime/src/time.cpp#L447)) and add these line of code there? (Only for future, for now this PR is fine with me)

pilz_utils/test/unittest_sleep.cpp
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright (c) 2019 Pilz GmbH & Co. KG
```
HiddenBug: 2020?

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
```diff
@@ -93,14 +112,32 @@ void PilzJointTrajectoryControllerTest::SetUp()
...
+
+  // Set up simulated time
+  ros::Time start_time{TIME_SIMULATION_START_SEC};
+  ros::Time::setNow(start_time);
+}
+
+void PilzJointTrajectoryControllerTest::startController()
+{
+  ros::Time current_time{ros::Time::now()};
+  controller_->starting(current_time);
```
HiddenBug, line 125: In function `PilzJointTrajectoryControllerTest::updateController()` you set:
`last_update_time_ = current_time;`
Should you not do the same here? Because otherwise, the first time `PilzJointTrajectoryControllerTest::updateController()` is called, it probably sets an incorrect time for last call to `update()` which, in my opinion, is the time since ` controller_->starting()` was called.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
```diff
@@ -93,14 +112,32 @@ void PilzJointTrajectoryControllerTest::SetUp()
...
+}
+
+void PilzJointTrajectoryControllerTest::updateController()
+{
+  ros::Time current_time{ros::Time::now()};
+  controller_->update(current_time, current_time - last_update_time_);
+  last_update_time_ = current_time;
 }
 
 ::testing::AssertionResult PilzJointTrajectoryControllerTest::waitForIsExecutingServiceResult(bool expectation)
```
HiddenBug: Function `PilzJointTrajectoryControllerTest::waitForIsExecutingServiceResult()` and function `PilzJointTrajectoryControllerTest::waitForIsExecutingResult` are nearly identical. Is it possible to use a function template, to avoid code duplication?

pilz_utils/include/pilz_utils/sleep.h
```diff
@@ -0,0 +1,49 @@
...
+
+#include <ros/ros.h>
+#include <ros/time.h>
+
+namespace pilz_utils
+{
+
+/**
+ * \brief Extends ros::Duration::sleep() by increasing the simulation time, if needed.
+ * \note The usage of ros::Time in this function is not thread-safe.
```
HiddenBug: Explain shorty why, it is not thread-safe. Or more precisely, explain what the user of the function should not do.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
```diff
@@ -206,10 +242,13 @@ TEST_F(PilzJointTrajectoryControllerTest, testIsExecutingServiceCallback)
...
   controller_->state_ = controller_->RUNNING;
 
-  controller_->update(ros::Time::now(), ros::Duration(STOP_TRAJECTORY_DURATION + SMALL_PERIOD));
+  updateController();
+  ros::Duration stop_duration{STOP_TRAJECTORY_DURATION + DEFAULT_UPDATE_PERIOD_SEC};
+  progressInTime(stop_duration);
+  updateController();
 
   EXPECT_TRUE(controller_->handleIsExecutingRequest(req, resp));
   EXPECT_FALSE(resp.success) << "Controller is executing unexpectedly";
```
HiddenBug: Sorry, but I don't understand these lines of code. Why should we wait the stop trajectory duration? Why should the controller execute a stop trajectory, after being started? Is this a negative test, to ensure that the controller does not do this?

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
```diff
@@ -181,8 +218,7 @@ TEST_F(PilzJointTrajectoryControllerTest, testInitializiation)
  *    7. is_executing service returns success=true after a while
  *    8. is_executing service returns success=false
  *    9. is_executing service returns success=true after a while
- *    10. is_executing service returns success=true
- *    11. is_executing service returns success=false
+ *    10. is_executing service returns success=false
  */
 TEST_F(PilzJointTrajectoryControllerTest, testIsExecutingServiceCallback)
```
HiddenBug: In my opinion this test is way to long and does to much in one go. I'd rather wait longer for several tests to execute then read to this one big test.

HiddenBug: The documentation of the test is extremely confusing. The documented steps do not separate the tests correctly. 
For example Step 9 and Step 10 belong together because the idea is to test that the controller stops a running trajectory if `hold` is called. The fact that this test part is separated by the to steps is extremely confusing and leads to a test which, in my opinion, is nearly impossible to maintain for developers who do not have infinite time.
Another example: 
Step 7 and step 8 also belong together, because we want to test that the controller returns `true` if a trajectory is executing and `false` if the trajectory is done. For me this is a logical unit and should not be separate in different steps. This is confusing.

HiddenBug: It even gets worse. There is another test named `testIsExecuting()` which basically does the same, except for a "minor" function difference. Can we not template the test, in order, to avoid such an extremely messy copy & paste situation.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
```diff
@@ -421,8 +473,7 @@ TEST_F(PilzJointTrajectoryControllerTest, testDoubleRequest)
  *    7. is_executing() returns true after a while
  *    8. is_executing() returns false
  *    9. is_executing() returns true after a while
- *    10. is_executing() returns true
- *    11. is_executing() returns false
+ *    10. is_executing() returns false
  */
 TEST_F(PilzJointTrajectoryControllerTest, testIsExecuting)
```
HiddenBug: There is another test named `testIsExecutingServiceCallback()` which basically does the same, except for a "minor" function difference. Can we not template the test, in order, to avoid such an extremely messy copy & paste situation.

pilz_utils/include/pilz_utils/sleep.h
HiddenBug: A general note:
Try to think from a user perspective: "What should the user of this function know?"

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
HiddenBug: An alternative to this change would be to replace the [time.cpp file](https://github.com/ros/roscpp_core/blob/kinetic-devel/rostime/src/time.cpp#L447) in the unit test compilation unit.

HiddenBug (commented):
Basically I'm ok with the changes you are proposing. 
I'm, however, not ok with the tests in `unittest_pilz_joint_trajectory_controller.cpp`. More precisely with the tests `testIsExecutingServiceCallback` and `testIsExecuting`. Please have a look at my comments and decide for yourself. I do not have the rights on this repository to "Request changes" with this review. But that is what I would choose as answer.

pilz_utils/include/pilz_utils/sleep.h
martiniil: Done.

pilz_utils/include/pilz_utils/sleep.h
martiniil: Done.

pilz_utils/include/pilz_utils/sleep.h
martiniil: The existing sleep function totally makes sense. To be honest I'm not really happy with my solution. I would rather overthink our tests.

I think using sim-time improves the unittest. However there are still open questions.

So no, I would not propose an upstream change.

pilz_utils/test/unittest_sleep.cpp
martiniil: Done

pilz_utils/include/pilz_utils/sleep.h
martiniil: Done

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
martiniil: I think you are right. Done.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
martiniil: See https://github.com/ros-controls/ros_controllers/blob/4e132cdd955f0b16ec32b936ff08757861c0bb55/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L150

I don't know why the robot should be moving when you start the controller. But anyways this ensures that it holds its position at the start.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
martiniil: See above.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
martiniil: See below.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
martiniil: @HiddenBug Thanks for your thorough review. I will dive into it and open a new PR. Can we merge this one in the mean-time?

martiniil: > @HiddenBug Thanks for your thorough review. I will dive into it and open a new PR. Can we merge this one in the mean-time?

Please ignore this comment. I will add more commits.

martiniil:
@HiddenBug I addressed your comments. Requesting second review.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
```diff
@@ -382,159 +260,334 @@ TEST_F(PilzJointTrajectoryControllerTest, testDoubleRequest)
...
+protected:
+  testing::AssertionResult invokeIsExecuting();
+  testing::AssertionResult waitForIsExecutingResult(bool expectation);
+
+  /**
+   * @brief Perform init, start, unhold and update, such that controller is ready for executing.
+   */
+  testing::AssertionResult performFullControllerStartup();
+
+  bool is_executing_result;
```
hslusarek, line 345: I'm not quite sure if I understand the need for this member.  In `invokeIsExecuting()` you are calling the currently used `is_executing_function` to check if the controller is executing. The corresponding function can return the controller state via`testing::AssertionSuccess` or `testing::AssertionFailure`. So why do you need to transfer and, consequently, check the executing state via the member `is_executing_result`?

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
```diff
@@ -382,159 +260,334 @@ TEST_F(PilzJointTrajectoryControllerTest, testDoubleRequest)
...
+{
+  std_srvs::TriggerRequest req;
+  std_srvs::TriggerResponse resp;
+  if(!controller->handleIsExecutingRequest(req, resp))
+  {
+    return testing::AssertionFailure() << "Callback of is_executing returned false unexpectedly.";
+  }
+
+  result = resp.success;
+  return testing::AssertionSuccess();
```
hslusarek, line 327: Why not:
```
return resp.success? testing::AssertionSuccess() : testing::AssertionFailure();
```
Is this not possible? Or you can also do:
```
if (resp.success)
{
    return testing::AssertionSuccess(); 
}
else
{
    return testing::AssertionFailure();
}
```

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
```diff
@@ -382,159 +260,334 @@ TEST_F(PilzJointTrajectoryControllerTest, testDoubleRequest)
...
+///////////////////////////////////////////////
+//    Parameterized tests for IsExecuting    //
+///////////////////////////////////////////////
+
+typedef std::function<testing::AssertionResult(const ControllerPtr&, bool&)> InvokeIsExecuting;
+
+static testing::AssertionResult InvokeIsExecutingMethod(const ControllerPtr& controller, bool& result)
+{
+  result = controller->is_executing();
+  return testing::AssertionSuccess();
```
hslusarek, line 314: Why not:
```
return controller->is_executing()? testing::AssertionSuccess() : testing::AssertionFailure();
```
Is this not possible? Or you can also do:
```
if (controller->is_executing())
{
    return testing::AssertionSuccess(); 
}
else
{
    return testing::AssertionFailure();
}
```

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
```diff
@@ -382,159 +260,334 @@ TEST_F(PilzJointTrajectoryControllerTest, testDoubleRequest)
...
+{
+  ASSERT_TRUE(controller_->init(hardware_, nh_, controller_nh_)) << "Failed to initialize the controller.";
+  controller_->state_ = controller_->INITIALIZED;
+
+  EXPECT_TRUE(invokeIsExecuting());
+  EXPECT_FALSE(is_executing_result) << "Controller is executing unexpectedly";
+}
+
+/**
+ * @brief Test the is_executing functionality of the started controller.
```
hslusarek: All test documentations are written as if we want to test the `is_executing` function. But in my opinion, we actually want to test the different controller functionalities whereas the `is_executing` function is just a mean to achieve this. Therefore, I suggest to describe what controller functionality the test is supposed to test instead.
What do you think?

**Please note:**
The tests are now much shorter. So, in my opinion, it is enough to provide a good short description of what the test wants to test and then skip/delete the detailed test-step descriptions.

**Please note:**
This comment holds for all test documentations in this test suite.

hslusarek (changes requested):
Only minor things. Just look at them, decide what you want to do and then merge.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
martiniil: I added a comment as an explanation:

> //! The return value indicates if the call was successful (in case of a service callback), the actual result
//! of the is-executing-function is assigned (via reference) to the second argument.

martiniil: In other words, this is used to distinguish different outcomes of the service callback.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
martiniil: See above.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
martiniil: See above.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
martiniil: Interesting. I had the opinion that we test the `is_executing` functionality. But it is a matter of point of view.

martiniil: I introduced short descriptions. Done.

martiniil: Will merge as soon as build checks are through.

PR #312, opened 2020-02-14
Commented modbus spec Closes #309

### LINKED ISSUE #309: Modbus Specification
Could you please briefly describe the modbus specification here?
Not only the registers themselves, but also the contents of the registers, which value stands for which function. Also if this function is already available and how to activate it would not be bad. 
For example;

- Yaml-Parameter: OPERATION_MODE
- Register: 970
- 1-->Manual reduced speed, 2-->Manual High speed, 3-->Automatik
- prbt_moveitconfig: arg name="has_operation_mode_support" default="false"/

- not available yet/work in progress/released

JonathanGruner:
Commented it with actual spec. 
Please check, if the information matches with your program.


ct2034:
Closes #309

prbt_hardware_support/config/modbus_read_api_spec_pnoz.yaml
```diff
@@ -1,5 +1,10 @@
 # PNOZMulti Software V2.1dev
-STO: 512
-VERSION: 513
-BRAKETEST_REQUEST: 514
-OPERATION_MODE: 515
\ No newline at end of file
+# Generic: Enabled the whole time
```
ct2034, line 2: The purpose of this file is to define, under which register the Modbus adapter can read the different values.
It should not document some behaviour of how and when those are enabled or disabled.
It would maybe make sense to use these lines to generally describe what the register content describes or skip them altogether.

```suggestion
# STO: This register contains the state of the robots emergency stop signal
```

prbt_hardware_support/config/modbus_read_api_spec_pnoz.yaml
```diff
@@ -1,5 +1,10 @@
...
+STO: 512 # robot is only allowed to move when the signal is 1
+VERSION: 513 # version of modbus register specification, actual one is: 2
+
+# Braketest: Disabled per default, can be enabled in robot.launch with parameter "has_braketest_support"
```
ct2034:
```suggestion
# Braketest: This is used to signalize that a brake test is requested by the safety controller
```

ct2034 (changes requested):

jschleicher:
@ct2034 Please just go ahead and implement your suggested changes for an easier review.

martiniil:
Continued in #385.

PR #313, opened 2020-02-14
Commented modbus_read_api_spec_pss4000.yaml for usability

JonathanGruner:
Commented it with actual spec. 
Please check, if the information matches with your program.


JonathanGruner:
I just changed STO with RUN_PERMITTED, because STO is a safety related signal. 
Furthermore the RUN_PERMITTED Signal appears 340 ms before the STO signal.
In fact "STO" is just wrong I would request an immediate change (I know that you have to change it in all the other files too)

PR #314, opened 2020-02-14
Mention pilz_templates in readme

JonathanGruner:
Some user didn't even know that there is such a package in our repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,10 +118,7 @@ then replace the pipeline in the above command by `pipeline:=pilz_command_planne
 If you've created an application package with your own launch file as described in the
 [tutorials](https://wiki.ros.org/pilz_robots/Tutorials/ModelYourApplicationWithPRBT#Create_your_application_ROS_package),
 you can easily adjust many other configuration parameters.
-See the template and comments in the [pilz_tutorials package](https://github.com/PilzDE/pilz_tutorials).
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -118,7 +118,7 @@ then replace the pipeline in the above command by `pipeline:=pilz_command_planne
 If you've created an application package with your own launch file as described in the
 [tutorials](https://wiki.ros.org/pilz_robots/Tutorials/ModelYourApplicationWithPRBT#Create_your_application_ROS_package),
 you can easily adjust many other configuration parameters.
-See the template and comments in the [pilz_tutorials package](https://github.com/PilzDE/pilz_tutorials).
```
ct2034, line 121: is this the right place for it?

ct2034 (changes requested):

ct2034:
I would suggest to mention the templates in a separate section in this readme

README.md
JonathanGruner: Yes! Because pilz_templates is an essential package, which is used by nearly every customer working with our robot. Because this readme is the main page and entry point, I would mention it there to avoid confusion.

JonathanGruner:
See comment above.

ct2034:
this is what i meant. @JonathanGruner

PR #296, opened 2020-01-09
Cartesian speed limitation in PJTC [Doc update]

martiniil:

PR #308, opened 2020-02-06
Feature/speed limit pjtc extend controller

hslusarek:
PR to "correctly" integrate Cartesian velocity check, into `PilzJointTrajectoryController` (and its base class `JointTrajectoryController`).

PR #318, opened 2020-02-19
Delete old speed observer

hslusarek:
Please note:
The operation mode option won't work after this PR because it waits for the `set_speed_limit` service which is no longer provided by somebody.
PR #296 will introduce the service `set_speed_limit` again. Then the controller will provide the service.


martiniil:
This change was added to #308.

PR #316, opened 2020-02-17
Changing references to `pilz_command_planner` into `pilz_industrial_motion_planner` in moveit

ct2034:
Follow-up after the merge of `pilz_trajectory_generation` into ros-planning/moveit as `pilz_industrial_motion_planner`.

See ros-planning/moveit#1893

Next steps:
- [x] close this p.r. once it is ready
- [x] create new branch noetic-devel and push it
- [x] add branch protection rule for noetic-devel

PR #319, opened 2020-03-09
add pipeline for merging the planner into moveit

jschleicher:
For the moment we have xml-files for both the old an the new pipeline in prbt_moveit_config. The old one will be replaced by a deprecation warning after the merge is complete; finally after a long enough migration time it can be removed, see #316.


ct2034 (approved):

ct2034:
There was something else necessary for the planner tests to succeed!
If travis now becomes green, we can merge it.

jschleicher:
Thank you for adding the `move_group.launch` file changes!

PR #321, opened 2020-03-11
enable ccache

jschleicher:
to speed up subsequent builds


ct2034 (approved):
this makes sense. I thought we did this already.

.travis.yml
```diff
@@ -4,6 +4,10 @@ sudo: required
 dist: trusty
 language: generic
 
+cache:
+  directories:
+    - $HOME/.ccache
+
```
ct2034, line 8:
```suggestion
cache: ccache
```
similar to https://github.com/PilzDE/pilz_industrial_motion/pull/254

ct2034 (changes requested):

.travis.yml
ct2034: according to https://docs.travis-ci.com/user/caching/#ccache-cache

jschleicher: @ct2034 the shorter form doesn't seem to work - see https://github.com/PilzDE/pilz_industrial_motion/pull/254#pullrequestreview-373562783

ct2034: that seems odd.

ct2034 (approved):

PR #323, opened 2020-03-16
Introduce goal_time_tolerance to PJTC function is_executing

martiniil:
Makes the check for a controller execution more consistent with the actual goal execution.
Analogue to 7a18f6a

### DIFF
--- a/pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
+++ b/pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
@@ -78,9 +78,10 @@ bool PilzJointTrajectoryController<SegmentImpl, HardwareInterface>::is_executing
   {
     auto uptime {JointTrajectoryController::time_data_.readFromRT()->uptime.toSec()};
     typename TrajectoryPerJoint::const_iterator segment_it = findSegment(curr_traj[i], uptime);
+    const auto& tolerances = segment_it->getTolerances();
     // Times that preceed the trajectory start time are ignored here, so is_executing() returns false
     // even if there is a current trajectory that will be executed in the future.
-    if (segment_it != curr_traj[i].end() && uptime <= segment_it->endTime())
+    if (segment_it != curr_traj[i].end() && uptime < segment_it->endTime() + tolerances.goal_time_tolerance)
     {
       is_executing = true;
       break;


pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
```diff
@@ -74,13 +74,16 @@ bool PilzJointTrajectoryController<SegmentImpl, HardwareInterface>::is_executing
...
+
   for (unsigned int i = 0; i < JointTrajectoryController::joints_.size(); ++i)
   {
     auto uptime {JointTrajectoryController::time_data_.readFromRT()->uptime.toSec()};
     typename TrajectoryPerJoint::const_iterator segment_it = findSegment(curr_traj[i], uptime);
+    const SegmentTolerancesPerJoint& tolerances = segment_it->getTolerances();
     // Times that preceed the trajectory start time are ignored here, so is_executing() returns false
     // even if there is a current trajectory that will be executed in the future.
-    if (segment_it != curr_traj[i].end() && uptime <= segment_it->endTime())
+    if (segment_it != curr_traj[i].end() && uptime < segment_it->endTime() + tolerances.goal_time_tolerance)
```
hslusarek, line 84: Are you aware of the fact, that not always a tolerance is stated. For example the stop trajectory, does not define a tolerance. In these cases the equation collapses to the original form, except for the `=`. Therefore, I suggest to still use `<=` instead of `<`. What do you think?

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
```diff
@@ -74,13 +74,16 @@ bool PilzJointTrajectoryController<SegmentImpl, HardwareInterface>::is_executing
 
   bool is_executing {false};
 
+  typedef joint_trajectory_controller::SegmentTolerancesPerJoint<typename Segment::Scalar> SegmentTolerancesPerJoint;
```
hslusarek: I suggest to use the "new" alias-decleration. See [nextptr.com](https://www.nextptr.com/tutorial/ta1193988140/how-cplusplus-using-or-aliasdeclaration-is-better-than-typedef)

Is it maybe even possible to skip the `typedef`if you use `auto`?

hslusarek (changes requested):
Only minor things, looks good to me. Look at my comments and decide if you want to change something and then simply merge.

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
martiniil: Using `auto` works.

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
martiniil: I would keep it analogue to the update function:

>       if (time_data.uptime.toSec() < segment_it->endTime())

https://github.com/ros-controls/ros_controllers/blob/b873e45c16c93b9960b106051d68cf3f225ad0ff/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L403

PR #331, opened 2020-04-07
Default Configuration per Safety Controller kinetic

ct2034:


prbt_hardware_support/README.md
```diff
@@ -16,6 +16,17 @@ The following table contains the supported configurations. The default configura
 
 For more on gripper models see also [prbt_grippers](https://github.com/PilzDE/prbt_grippers).
 
+# Default Configuration per Safety Controller
```
hslusarek: I know what this section is all about, but even for me it is difficult to differentiate between the section "_Supported configurations_" and the section "_Default Configuration per Safety Controller_". Therefore, I suggest the following changes:

Firstly, the changes for section "_Supported configurations_":
- Reorder the column of table the following way: "Parameter Description", "Argument name in `prbt_support/robot.launch`" and then "Option values in `prbt_support/robot.launch`". Skip column "Options", these are just redundant information.
- Rename "Option values in `prbt_support/robot.launch`" to "Possible values in `prbt_support/robot.launch`".
- Instead of `empty_string` for "no gripper", we should support `none` it is easier to understand and needs no further explanation. For backward compatibility, we can silently still support `empty_string`. We, however, should no longer advertise it. Please make the necessary changes.
- Change the text above the table the following way:
"The following table shows the names and the possible argument values of the features/setups supported by our hardware. The features/setups can be configured via [`prbt_support/robot.launch`](https://github.com/PilzDE/pilz_robots/blob/melodic-devel/prbt_support/launch/robot.launch)."
- Rename the section to something like: "_How-to activate/de-activate features/setups_".

Secondly, the changes for section "_Default Configuration per Safety Controller_"
- Change the text above the table to something like: 
"Currently, we only support the following configurations depending on the safety controller. If the safety controller is changed, please ensure that  the arguments are set as shown in the table."
- Remove the section title "_Default Configuration per Safety Controller_" and let the text of this section and the table of this section directly follow the table and section of the previous one.

hslusarek (changes requested):
Review comments.

prbt_hardware_support/README.md
ct2034: Changed according to our phone call

Issue #333, opened 2020-04-08
Introduce timeout in AsyncTest


martiniil:
**Problem**
Regard a test that waits for an asynchronous event using `AsyncTest::barricade`. If the event is not triggered the test waits until the time for the test is completely used up. Then you cannot deduce from the test report why it failed.
This makes it very hard to debug.

**Possible solution**
Add a timeout to `AsyncTest::barricade` such that a test failure with a context-related message can be generated.

PR #338, opened 2020-04-20
Add timeout to AsyncTest::barricade()

martiniil:
See #333

### DIFF
--- a/pilz_testutils/include/pilz_testutils/async_test.h
+++ b/pilz_testutils/include/pilz_testutils/async_test.h
@@ -75,20 +75,20 @@ class AsyncTest
 
     /**
      * @brief Will block until the event given by clear_event is triggered or a timeout is reached.
-     * Unblocks immediately, if the event was triggered in advance.
+     * Unblocks immediately, if the event was on the waitlist.
      *
      * @param clear_event Event that allows the test to pass on
-     * @param timeout_ms Timeout [ms]. If negative, no timeout is used.
+     * @param timeout_ms Timeout [ms] (optional).
      * @return True if the event was triggered, false otherwise.
      */
     bool barricade(const std::string &clear_event, const int timeout_ms = -1);
 
     /**
      * @brief Will block until all events given by clear_events are triggered or a timeout is reached.
-     * Events triggered in advance take effect, too.
+     * Events on the waitlist are taken into account, too.
      *
      * @param clear_events List of events that allow the test to pass on
-     * @param timeout_ms Timeout [ms]. If negative, no timeout is used.
+     * @param timeout_ms Timeout [ms] (optional).
      * @return True if all events were triggered, false otherwise.
      */
     bool barricade(std::initializer_list<std::string> clear_events, const int timeout_ms = -1);
--- a/pilz_testutils/include/pilz_testutils/async_test.h
+++ b/pilz_testutils/include/pilz_testutils/async_test.h
@@ -56,8 +56,9 @@ namespace testing
  * TEST_F(MyTest, testCase)
  * {
  *     EXPECT_CALL(*this, myMethod()).Times(1).WillOnce(ACTION_OPEN_BARRIER_VOID("myMethod"));
+ *     const timeout_ms {100};
  *     asyncCall(std::bind(&MyTest::myMethod, this));
- *     BARRIER("myMethod");
+ *     BARRIER("myMethod", timeout_ms) << "Timed-out waiting for myMethod call.";
  * }
  * \endcode
  */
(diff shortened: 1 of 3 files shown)


pilz_testutils/include/pilz_testutils/async_test.h
```diff
@@ -73,20 +74,24 @@ class AsyncTest
     void triggerClearEvent(const std::string &event);
 
     /**
-     * @brief Will block until the event given by clear_event is triggered. Unblocks immediately, if the event was
-     * triggered in advance.
+     * @brief Will block until the event given by clear_event is triggered or a timeout is reached.
+     * Unblocks immediately, if the event was triggered in advance.
      *
      * @param clear_event Event that allows the test to pass on
+     * @param timeout_ms Timeout [ms]. If negative, no timeout is used.
```
ct2034: I would say, the timeout beeing negative is an internal thing. Coul we not just say, it is an optional parameter. Then, users of the method just not set it, if they want no timeout.
```suggestion
     * @param timeout_ms Timeout [ms]. (optional)
```

pilz_testutils/include/pilz_testutils/async_test.h
```diff
@@ -73,20 +74,24 @@ class AsyncTest
     void triggerClearEvent(const std::string &event);
 
     /**
-     * @brief Will block until the event given by clear_event is triggered. Unblocks immediately, if the event was
-     * triggered in advance.
+     * @brief Will block until the event given by clear_event is triggered or a timeout is reached.
+     * Unblocks immediately, if the event was triggered in advance.
```
ct2034:
```suggestion
     * Unblocks immediately, if the event was triggered before the timeout.
```
I would prefer this wording. `in advance` sounds to me like being triggered before the barricade is defined. Which of cause makes no sense. But I think it is clearer like this

pilz_testutils/include/pilz_testutils/async_test.h
```diff
@@ -73,20 +74,24 @@ class AsyncTest
...
+     * @return True if the event was triggered, false otherwise.
      */
-    void barricade(const std::string &clear_event);
+    bool barricade(const std::string &clear_event, const int timeout_ms = -1);
 
     /**
-     * @brief Will block until all events given by clear_events are triggered. Events triggered in advance take effect,
-     * too.
+     * @brief Will block until all events given by clear_events are triggered or a timeout is reached.
+     * Events triggered in advance take effect, too.
```
ct2034: what effect?

pilz_testutils/include/pilz_testutils/async_test.h
```diff
@@ -73,20 +74,24 @@ class AsyncTest
...
      *
      * @param clear_events List of events that allow the test to pass on
+     * @param timeout_ms Timeout [ms]. If negative, no timeout is used.
```
ct2034: see above comment

pilz_testutils/src/async_test.cpp
```diff
@@ -43,10 +44,20 @@ void AsyncTest::barricade(std::initializer_list<std::string> clear_events)
...
   {
-    cv_.wait(lk);
+    if (timeout_ms < 0)
+    {
+      cv_.wait(lk);
+    }
+    else if (cv_.wait_for(lk, end_time_point - std::chrono::system_clock::now()) == std::cv_status::timeout)
+    {
+      return clear_events_.empty();
+    }
```
ct2034: I think it would be more readlable:
```suggestion
    else 
    {
      std::future_status status = cv_.wait_for(lk, end_time_point - std::chrono::system_clock::now())
      if (status == std::cv_status::timeout)
      {
        return clear_events_.empty();
      }
    }
```

ct2034 (changes requested):
Just some small remarks

pilz_testutils/include/pilz_testutils/async_test.h
martiniil: > sounds to me like being triggered before the barricade is defined

That is exactly what it means.

pilz_testutils/include/pilz_testutils/async_test.h
martiniil: Thanks. Sounds good.

pilz_testutils/include/pilz_testutils/async_test.h
martiniil: However I tried to come up with a better explanation.

pilz_testutils/src/async_test.cpp
martiniil: Thanks. I would have committed it, but you accidently used `future_status` instead of `cv_status` ...

pilz_testutils/include/pilz_testutils/async_test.h
ct2034: yes, thats good!

ct2034 (approved):
I approve it like this. 
But just one thought: Can it cause problems that it now has a return value and previously not. https://github.com/PilzDE/pilz_robots/pull/338/files#diff-b322bf0080db26bcd03771677d96d84eR84
Feel free to merge, though

PR #360, opened 2020-04-30
add missing run depend on status_indicator

jschleicher:
According to roslaunch-check: [/tmp/ws/src/pilz_robots/prbt_support/launch/robot.launch]:
>	Missing package dependencies: prbt_hardware_support/package.xml: pilz_status_indicator_rqt

See test failure on [buildfarm](http://build.ros.org/job/Mdev__pilz_robots__ubuntu_bionic_amd64/126/console)


martiniil:
This results in a circular dependency ... I will try to solve it.

**melodic-devel**
rospackgraph






**fix/depend-status-indicator**
rospackgraph_broken

PR #362, opened 2020-05-01
Fix/circular exec depend

martiniil:
Replacement for #360, which introduced a circular dependency between `prbt_hardware_support` and `pilz_status_indicator_rqt`.

Should also fix failures on ROS buildfarm mentioned in #360.


jschleicher:
Why does `pilz_status_indicator_rqt` depend on `*_hardware_support`? I can only find `OperationMode` message, right? If that's the only usage I'd suggest to move the OperationMode message into `pilz_msgs` - it's common practice to put messages that are used outside of the own package there, because of such dependencies.

jschleicher:
Your PR still looks valid, to keep separate things seperate - indepently of the above question.

agutenkunst (approved):
Looks good!

PR #378, opened 2020-05-06
Extend stop1 acceptance test

martiniil:
The new tests verify that #93 is solved.

### DIFF
--- a/prbt_hardware_support/test/acceptancetests/acceptancetest_stop1.md
+++ b/prbt_hardware_support/test/acceptancetests/acceptancetest_stop1.md
@@ -58,8 +58,8 @@ These acceptance tests check that the real robot system reacts to RUN_PERMITTED 
 ## 4. Fast RUN_PERMITTED switching
 ### Test Sequence
   1. Same as Test 1
-  2. Perform as fast as you can: Release and press acknowledge button. Repeat several times.
-  3. Perform as fast as you can: Press and release acknowledge button. Repeat several times.
+  2. Perform the sequence "Release acknowledge button - press acknowledge button" very quickly. Repeat this test multiple times.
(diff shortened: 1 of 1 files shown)


prbt_hardware_support/test/acceptancetests/acceptancetest_stop1.md
```diff
@@ -54,3 +54,14 @@ These acceptance tests check that the real robot system reacts to RUN_PERMITTED
 ### Expected Results
   1. Same as Test 1
   2. Robot should stop immediatly and smoothly.
+
+## 4. Fast RUN_PERMITTED switching
+### Test Sequence
+  1. Same as Test 1
+  2. Perform as fast as you can: Release and press acknowledge button. Repeat several times.
```
SansoneG: I don't see how Point 2 and 3 are different.

martiniil: The order is different. 2. is about the sequence release->press and 3. the other way around.

SansoneG: I think I understand now. The wording might be confusing. I interpreted "repeat several times" as in "keep on pressing and releasing the button as fast as you can", but I now think it is meant as "repeat the single action of a fast release+press multiple times no matter how much time passes in between the single actions."

martiniil: Is it better now?

SansoneG (approved):

prbt_hardware_support/test/acceptancetests/acceptancetest_stop1.md
SansoneG: Yes, thank you!

PR #384, opened 2020-05-08
Enhance safety interface launchfiles

martiniil:
* brake-test launch files are only used via `safety_interface.launch`, so loading the api-spec is redundant.
* Add argument `read_api_spec_file` to `modbus_client.launch`, the default is unchanged.
* Fix indentation in `modbus_client.launch`

### DIFF
--- a/prbt_hardware_support/launch/modbus_client.launch
+++ b/prbt_hardware_support/launch/modbus_client.launch
@@ -28,8 +28,9 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 <!-- define the connected safety hardware -->
 <arg name="safety_hw" default="pss4000" />
+<arg name="read_api_spec_file" default="$(find prbt_hardware_support)/config/modbus_read_api_spec_$(arg safety_hw).yaml" />
 
-<rosparam ns="/prbt/read_api_spec" command="load" file="$(find prbt_hardware_support)/config/modbus_read_api_spec_$(arg safety_hw).yaml" />
+<rosparam ns="/prbt/read_api_spec" command="load" file="$(arg read_api_spec_file)" />
   <arg name="has_register_range_parameters"
        value="$(eval arg('index_of_first_register_to_read')!='' and arg('num_registers_to_read')!='')" />
 
--- a/prbt_hardware_support/launch/modbus_client.launch
+++ b/prbt_hardware_support/launch/modbus_client.launch
@@ -17,20 +17,20 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 <launch>
 
-<!-- default modbus server ip (PSS400) and port number -->
-<arg name="modbus_server_ip" default="192.168.0.10" />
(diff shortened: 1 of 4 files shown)


jschleicher (commented):
I cannot find the first line, setting `/prbt/write_api_spec`. Could you please point me to the redundant file?

martiniil:
> I cannot find the first line, setting `/prbt/write_api_spec`. Could you please point me to the redundant file?

https://github.com/PilzDE/pilz_robots/blob/7adc6aa37662d258cddaf7d6bb11a9aff57b1a2e/prbt_hardware_support/launch/safety_interface.launch#L32

Did you search for this line?

prbt_hardware_support/launch/modbus_client.launch
```diff
@@ -17,19 +17,20 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+  <arg name="num_registers_to_read" default="" />
 
-<!-- define the connected safety hardware -->
-<arg name="safety_hw" default="pss4000" />
+  <!-- define the connected safety hardware -->
+  <arg name="safety_hw" default="pss4000" />
+  <arg name="read_api_spec_file" default="$(find prbt_hardware_support)/config/modbus_read_api_spec_$(arg safety_hw).yaml" />
 
-<rosparam ns="/prbt/read_api_spec" command="load" file="$(find prbt_hardware_support)/config/modbus_read_api_spec_$(arg safety_hw).yaml" />
+  <rosparam ns="/prbt/read_api_spec" command="load" file="$(arg read_api_spec_file)" />
```
jschleicher, line 33: then this line would be redundant as well??

jschleicher (commented):
Ok, thanks! So we drop the "write" argument saying it's unused while introducing a new unused "read" argument, correct? If that's the case I'd suggest to defer the discussion until we have an architectural overview that defines the public interfaces - tbd when we write the modbus tutorial.
We could now merge 7adc6aa37662d258cddaf7d6bb11a9aff57b1a2e.

prbt_hardware_support/launch/modbus_client.launch
martiniil: I think it is not redundant because we mention `modbus_client.launch` as a standalone launch file in the ReadMe.

martiniil:
> 
> 
> Ok, thanks! So we drop the "write" argument saying it's unused while introducing a new unused "read" argument, correct? If that's the case I'd suggest to defer the discussion until we have an architectural overview that defines the public interfaces - tbd when we write the modbus tutorial.
> We could now merge [7adc6aa](https://github.com/PilzDE/pilz_robots/commit/7adc6aa37662d258cddaf7d6bb11a9aff57b1a2e).

Thanks for your review.

The `read_api_spec` argument is already present in `safety_interface.launch`, so I thought it would be more clean to pass it to `modbus_client.launch` instead of defining the filename twice. I don`t see that it is unused?

agutenkunst (approved):
Thank you!

PR #406, opened 2020-06-08
Limit joint acceleration in pilz_joint_trajectory_controller

martiniil:
Based on #308
First commit here: 6e4d18a


pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
```diff
@@ -63,6 +107,9 @@ bool PilzJointTrajectoryController<SegmentImpl, HardwareInterface>::init(Hardwar
 {
   bool res = JointTrajectoryController::init(hw, root_nh, controller_nh);
 
+  ros::NodeHandle limits_nh(controller_nh, "limits");
```
SansoneG: Magic String "limits"

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
```diff
@@ -225,6 +272,33 @@ inline void PilzJointTrajectoryController<SegmentImpl, HardwareInterface>::updat
...
+
+template <class SegmentImpl, class HardwareInterface>
+inline bool PilzJointTrajectoryController<SegmentImpl, HardwareInterface>::isPlannedJointAccelerationOK(
+    const ros::Duration& period) const
+{
+  for (unsigned int i = 0; i < JointTrajectoryController::getNumberOfJoints(); ++i)
+  {
+    const double& old_velocity = JointTrajectoryController::old_desired_state_.velocity.at(i);
+    const double& new_velocity = JointTrajectoryController::desired_state_.velocity.at(i);
+    const double& acceleration = (new_velocity - old_velocity) / period.toSec();
```
SansoneG: What about negative acceleration? (e.g. old_velocity > new_velocity)

Should this value be absolute?

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
```diff
@@ -225,6 +272,33 @@ inline void PilzJointTrajectoryController<SegmentImpl, HardwareInterface>::updat
...
+  return isPlannedJointAccelerationOK(period) && isPlannedCartesianVelocityOK(period);
+}
+
+template <class SegmentImpl, class HardwareInterface>
+inline bool PilzJointTrajectoryController<SegmentImpl, HardwareInterface>::isPlannedJointAccelerationOK(
+    const ros::Duration& period) const
+{
+  for (unsigned int i = 0; i < JointTrajectoryController::getNumberOfJoints(); ++i)
+  {
+    const double& old_velocity = JointTrajectoryController::old_desired_state_.velocity.at(i);
```
SansoneG: We should consider not using simple double variables for physics calculations. Instead introduce velocity and acceleration types to avoid making errors with e.g. unit conversions.

pilz_control/test/acceptance_test_acceleration_limit.py
```diff
@@ -0,0 +1,125 @@
...
+from control_msgs.msg import *
+from trajectory_msgs.msg import JointTrajectory, JointTrajectoryPoint
+
+COMMAND_TOPIC_NAME = '/prbt/manipulator_joint_trajectory_controller/command'
+FOLLOW_JOINT_TRAJ_ACTION_NAME = '/prbt/manipulator_joint_trajectory_controller/follow_joint_trajectory'
+JOINT_NAMES = ['prbt_joint_1', 'prbt_joint_2', 'prbt_joint_3', 'prbt_joint_4', 'prbt_joint_5', 'prbt_joint_6']
+
+DEFAULT_QUEUE_SIZE = 10
+DEFAULT_TRAJECTORY_DURATION_S = 5
+WAIT_FOR_SERVICE_TIMEOUT_S = 10
```
SansoneG: why not make this a `rospy.Duration` from the start?

pilz_control/test/acceptance_test_acceleration_limit.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def __init__(self):
+        self._pub = rospy.Publisher(COMMAND_TOPIC_NAME, JointTrajectory, queue_size=DEFAULT_QUEUE_SIZE)
+
+        self._client = actionlib.SimpleActionClient(FOLLOW_JOINT_TRAJ_ACTION_NAME, FollowJointTrajectoryAction)
+
+        timeout = rospy.Duration(WAIT_FOR_SERVICE_TIMEOUT_S)
+        self._client.wait_for_server(timeout)
+
+    def sendMessage(self, position=[0.0]*len(JOINT_NAMES), velocity=[], time_from_start = DEFAULT_TRAJECTORY_DURATION_S):
```
SansoneG: having [0, 0, 0, 0, 0, 0] as the default position could be confusing. Might be better as required argument rather than optional.

pilz_control/test/acceptance_test_acceleration_limit.py
```diff
@@ -0,0 +1,125 @@
...
+        point = JointTrajectoryPoint()
+        point.positions = position
+        point.velocities = velocity
+        point.time_from_start = rospy.Duration(time_from_start)
+
+        traj.points.append(point)
+
+        self._pub.publish(traj)
+
+    def sendActionGoal(self, position=[0.0]*len(JOINT_NAMES), velocity=[], velocity_tolerance=0.0,
```
SansoneG: see line 42

pilz_control/test/acceptance_test_acceleration_limit.py
```diff
@@ -0,0 +1,125 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import actionlib
+import rospy
+
+from actionlib_msgs.msg import *
```
SansoneG: seems to me like no `actionlib_msg` objects are actually used

pilz_control/test/acceptance_test_acceleration_limit.py
```diff
@@ -0,0 +1,125 @@
...
+from control_msgs.msg import *
```
SansoneG: not all imports from `control_msgs` are used

pilz_control/test/acceptance_test_acceleration_limit.py
```diff
@@ -0,0 +1,125 @@
...
+            tol.name = joint_name
+            tol.velocity = velocity_tolerance
+            goal.path_tolerance.append(tol)
+
+        self._client.send_goal_and_wait(goal)
+
+
+def askForPermission(test_case):
+    s = raw_input('Perform ' + test_case + ' [(y)es, (n)o]?: ')
+    return (s == "y")
```
SansoneG: consider also accepting "Y"

pilz_control/test/acceptance_test_acceleration_limit.py
```diff
@@ -0,0 +1,125 @@
...
+
+def program():
+    trajectory_dispatcher = SinglePointTrajectoryDispatcher()
+
+    print('!!! BE CAREFUL. ROBOT MIGHT CRASH. !!!')
+
+    rospy.sleep(3.0)
+
+    print('Move to home position...')
+    trajectory_dispatcher.sendActionGoal()
```
SansoneG: This is for example what I meant by confusing in line 42. No parameters -> home position.

pilz_control/test/acceptance_test_acceleration_limit.py
```diff
@@ -0,0 +1,125 @@
...
+
+    start_position = [0.0, -0.5, 0.0, 0.0, 0.0, 0.0]
+    target_position = [0.0, 0.5, 0.0, 0.0, 0.0, 0.0]
+
+    if askForPermission('critical movement by topic'):
```
SansoneG: consider adding else statement with a `print("skipping")`

pilz_control/test/config/joint_limits.yaml
```diff
@@ -0,0 +1,40 @@
...
+    max_acceleration: 3.49
+  joint_with_undefined_max_acc:
+    has_velocity_limits: true
+    max_velocity: 1.57
+    has_acceleration_limits: true
+  joint_with_has_acc_lim_false:
+    has_velocity_limits: true
+    max_velocity: 1.57
+    has_acceleration_limits: false
+  joint_with_undefined_has_acc_lim:
```
SansoneG: Joint Name doesn't explicitely state that `max_acceleration` is also undefined.

pilz_control/test/integrationtest_pilz_joint_trajectory_controller.py
```diff
@@ -182,6 +185,20 @@ def wait_for_end_of_stop_observation(self):
             return self._stop_trajectory_ok
 
 
+    def _try_to_get_actual_position(self):
+        with self._controller_state_lock:
+            return self._actual_position
```
SansoneG: Not really part of this PR, but why is this variable called `_actual_position`? Considering the context it seems that `_position` would do the trick.

pilz_control/test/integrationtest_pilz_joint_trajectory_controller.py
```diff
@@ -182,6 +185,20 @@ def wait_for_end_of_stop_observation(self):
             return self._stop_trajectory_ok
 
 
+    def _try_to_get_actual_position(self):
```
SansoneG: Function should have docstring stating use and return value

pilz_control/test/integrationtest_pilz_joint_trajectory_controller.py
```diff
@@ -182,6 +185,20 @@ def wait_for_end_of_stop_observation(self):
...
+        with self._controller_state_lock:
+            return self._actual_position
+
+    def get_actual_position(self, timeout):
```
SansoneG: Function should include return value and exceptions thrown in docstring

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
```diff
@@ -27,6 +27,50 @@ static constexpr double SPEED_LIMIT_NOT_ACTIVATED{ -1.0 };
...
+ *
+ * Function assumes parameter server naming prefix '/joint_limits/' for Joint Names.
+ *
+ * @param nh NodeHandle to access parameter server.
+ * @param joint_names Vector of Strings for all joint names to get the acceleration limits for.
+ * @return std::vector<double> Requested acceleration limits as vector. Has the same length as param 'joint_names'.
+ *
+ * @throw InvalidParameterException Requested values not found on param server.
+ */
+std::vector<double> getJointAccelerationLimits(const ros::NodeHandle& nh, const std::vector<std::string> joint_names)
```
hslusarek: Due to the fact that this function is not supposed to be used outside of our `PilzJointTrajectoryController` I suggest to make this function `static`, to avoid future name collisions and so on. In case the function is supposed to be used outside the `PilzJointTrajectoryController`, I suggest to separate this function into a separate file.

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
```diff
@@ -27,6 +27,50 @@ static constexpr double SPEED_LIMIT_NOT_ACTIVATED{ -1.0 };
...
+{
```
hslusarek, line 62: The argument `joint_names` should be passed in as constant reference and not as constant copy.

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
```diff
@@ -27,6 +27,50 @@ static constexpr double SPEED_LIMIT_NOT_ACTIVATED{ -1.0 };
...
+    std::string param_name_to_read = joint_limits_naming_prefix + joint_names.at(i) + "/has_acceleration_limits";
+    if (!nh.getParam(param_name_to_read, has_acceleration_limits))
+    {
+      throw ros::InvalidParameterException("Failed to get the has_acceleration_limits for " + joint_names.at(i) +
+                                           " under param name >" + param_name_to_read + "<.");
+    }
+
+    if (!has_acceleration_limits)
+    {
+      acc_limits.at(i) = 0.;
```
hslusarek: When constructing the vector simply write:
```
std::vector<double> acc_limits(joint_names.size(), 0.);
```
Then you can delete this line. For more infos about the possible vector constructors see [cppreference.com](https://en.cppreference.com/w/cpp/container/vector/vector)

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
```diff
@@ -27,6 +27,50 @@ static constexpr double SPEED_LIMIT_NOT_ACTIVATED{ -1.0 };
...
+ */
+std::vector<double> getJointAccelerationLimits(const ros::NodeHandle& nh, const std::vector<std::string> joint_names)
+{
+  const std::string joint_limits_naming_prefix{ "/joint_limits/" };
+
+  std::vector<double> acc_limits(joint_names.size());
+  for (unsigned int i = 0; i < joint_names.size(); ++i)
+  {
+    bool has_acceleration_limits = false;
+    std::string param_name_to_read = joint_limits_naming_prefix + joint_names.at(i) + "/has_acceleration_limits";
```
hslusarek: The string creation and the string creation some lines below could be separated into a separate function for cleaner code.

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
```diff
@@ -27,6 +27,50 @@ static constexpr double SPEED_LIMIT_NOT_ACTIVATED{ -1.0 };
...
+    if (!nh.getParam(param_name_to_read, has_acceleration_limits))
+    {
+      throw ros::InvalidParameterException("Failed to get the has_acceleration_limits for " + joint_names.at(i) +
+                                           " under param name >" + param_name_to_read + "<.");
+    }
+
+    if (!has_acceleration_limits)
```
hslusarek: Separate the parameter creation of `param_name_to_read` in a separate function and then add the inverse of this condition to the condition `if (!nh.getParam(param_name_to_read, acc_limits.at(i)))` some lines below.

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
hslusarek: Yes, please, please, please. Vielleicht können wir sowas hier machen:
https://en.cppreference.com/w/cpp/language/user_literal

pilz_control/include/pilz_control/pilz_joint_trajectory_controller.h
```diff
@@ -216,6 +234,9 @@ class PilzJointTrajectoryController
   //! The currently max allowed speed for each frame on the Cartesian trajectory.
   std::atomic<double> cartesian_speed_limit_{ 0.0 };
 
+  //! The max allowed acceleration for each joint.
+  std::vector<double> acceleration_joint_limits_;
```
hslusarek: I'm actually not really happy with the data type choice here. Acceleration limit `0.0` is supposed to express that no acceleration limits are checked for the corresponding joint. This is bad style. I suggest that we use `boost::optional<double>` as data type (for more info about the optional data type, see [boost.org](https://www.boost.org/doc/libs/1_73_0/libs/optional/doc/html/index.html)). This allows us to express clearer, that  for some joints no acceleration limits are defined.

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
```diff
@@ -225,6 +272,33 @@ inline void PilzJointTrajectoryController<SegmentImpl, HardwareInterface>::updat
...
+template <class SegmentImpl, class HardwareInterface>
+inline bool PilzJointTrajectoryController<SegmentImpl, HardwareInterface>::isPlannedJointAccelerationOK(
+    const ros::Duration& period) const
+{
+  for (unsigned int i = 0; i < JointTrajectoryController::getNumberOfJoints(); ++i)
+  {
+    const double& old_velocity = JointTrajectoryController::old_desired_state_.velocity.at(i);
+    const double& new_velocity = JointTrajectoryController::desired_state_.velocity.at(i);
+    const double& acceleration = (new_velocity - old_velocity) / period.toSec();
+    if ((acceleration_joint_limits_.at(i) > 0.0) && (acceleration > acceleration_joint_limits_.at(i)))
```
hslusarek: Please choose another data type for the vector like described by me above, because
```
(acceleration_joint_limits_.at(i) > 0.0)
```
to express that no acceleration limits hold for a joint is bad style in 2020.

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
```diff
@@ -225,6 +272,33 @@ inline void PilzJointTrajectoryController<SegmentImpl, HardwareInterface>::updat
...
+    {
+      ROS_ERROR_STREAM("Acceleration limit violated by joint "
```
hslusarek: I suggest to use a named error log `ROS_ERROR_STREAM_NAMED(this->name_, MY_MSG)`, like mostly used in the base class.

hslusarek: Even better: Skip the error message because this code is supposed to be called in a real time function :)

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
hslusarek: Very good question. What about deceleration in general? Where these things considered properly?

Here properly [`std::abs()`](https://en.cppreference.com/w/cpp/numeric/math/fabs) does the trick?

hslusarek: You could write a small helper function for the acceleration calculation, to obtain cleaner code.

pilz_control/spec.dox
```diff
@@ -47,10 +47,14 @@
  *
  * @subsubsection speed_monitoring Speed Monitoring
  *
- * @definereq{Monitor_Speed_of_all_links_until_TCP}
+ * @definereq{Monitor_speed_of_all_links_until_TCP}
  *  - The system monitors all links of the robot (including the TCP, but not the end effector)
  *    in order to determine their translational speed.
  *
+ * @definereq{Monitor_joint_accelerations}
+ *  - The system monitors all joints of the robot and measures their angular accelerations.
```
hslusarek, line 55: What about the decelerations?

pilz_control/include/pilz_control/pilz_joint_trajectory_controller.h
```diff
@@ -17,9 +17,10 @@
 #ifndef PILZ_CONTROL_PILZ_JOINT_TRAJECTORY_CONTROLLER_H
 #define PILZ_CONTROL_PILZ_JOINT_TRAJECTORY_CONTROLLER_H
 
-#include <mutex>
-#include <memory>
 #include <atomic>
+#include <memory>
+#include <mutex>
+#include <vector>
```
hslusarek, line 24: `#include <string>` missing.

pilz_control/test/acceptance_test_acceleration_limit.py
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/env python
+# Copyright (c) 2019 Pilz GmbH & Co. KG
```
hslusarek: Incorrect year.

pilz_control/test/acceptance_test_acceleration_limit.py
```diff
@@ -0,0 +1,125 @@
...
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
```
hslusarek, line 16: This file does not fully comply with the PEP 8 Standard. Please ensure that the PEP 8 Standard is fulfilled.

pilz_control/test/acceptance_test_acceleration_limit.py
```diff
@@ -0,0 +1,125 @@
...
+        for joint_name in JOINT_NAMES:
+            tol = JointTolerance()
+            tol.name = joint_name
+            tol.velocity = velocity_tolerance
+            goal.path_tolerance.append(tol)
+
+        self._client.send_goal_and_wait(goal)
+
+
+def askForPermission(test_case):
```
hslusarek: Function name should be lower case.

pilz_control/test/acceptance_test_acceleration_limit.py
hslusarek: Function name should be lower case.

pilz_control/test/acceptance_test_acceleration_limit.py
```diff
@@ -0,0 +1,125 @@
...
+
+COMMAND_TOPIC_NAME = '/prbt/manipulator_joint_trajectory_controller/command'
+FOLLOW_JOINT_TRAJ_ACTION_NAME = '/prbt/manipulator_joint_trajectory_controller/follow_joint_trajectory'
+JOINT_NAMES = ['prbt_joint_1', 'prbt_joint_2', 'prbt_joint_3', 'prbt_joint_4', 'prbt_joint_5', 'prbt_joint_6']
+
+DEFAULT_QUEUE_SIZE = 10
+DEFAULT_TRAJECTORY_DURATION_S = 5
+WAIT_FOR_SERVICE_TIMEOUT_S = 10
+
+class SinglePointTrajectoryDispatcher:
```
hslusarek: It seems to me that in large parts class `SinglePointTrajectoryDispatcher` is a duplicate of the class `SinglePointTrajectoryDispatcher` which can be found in `acceptance_test_speed_monitoring.py`. Please remove the code duplication.

hslusarek (changes requested):
Code review (tests are reviewed separately).

pilz_control/test/integrationtest_pilz_joint_trajectory_controller.py
hslusarek: > unction should have docstring stating use and return value

A good reference literature on how to document code can be found in "Clean Code - A Handbook of Agile Software Craftmanship" by Robert C. Martin (Prentice Hall, 2008).  Have a look in chapter 4 which is about "Comments".
Here are some quotes from this book (Decide for yourself if the function really needs a docstring.):
- p.54: "Truth can only be found in one place: the code. [...] It is the only source of truly accurate information. [....]"
- p.70: "Short functions don't need much description. A well-chosen name for small function that does one thing is usually better than a comment header."
- p.71: "As useful as javadocs are for public APIs, they are anathema to code that is not intended for public consumption. [...]"

pilz_control/test/integrationtest_pilz_joint_trajectory_controller.py
hslusarek: > Function should include return value and exceptions thrown in docstring

See my comment above.

pilz_control/test/pjtc_test_helper.h
```diff
@@ -155,6 +135,55 @@ static bool updateUntilRobotMotion(RobotDriver* robot_driver,
...
+ * These deltas are added to the current position of the robot, to always move relatively and judge the size and speed
+ * of motion correctly.
+ */
+template <class RobotDriver>
+static GoalType generateSimpleGoal(RobotDriver* robot_driver,
+                                   const ros::Duration& goal_duration = ros::Duration(DEFAULT_GOAL_DURATION_SEC),
+                                   const float size_multiplier = 1)
+{
+  updateUntilNoRobotMotion<RobotDriver>(robot_driver);
+  std::vector<double> joint_positions = getMockRobotJointPositions<RobotDriver>(robot_driver);
```
hslusarek: Can be `const`.

pilz_control/test/pjtc_test_helper.h
```diff
@@ -155,6 +135,55 @@ static bool updateUntilRobotMotion(RobotDriver* robot_driver,
...
+  return joint_positions;
+}
+
+/**
+ * @brief Cycle through four different postion deltas to always get a motion, even if last goal fails.
+ * These deltas are added to the current position of the robot, to always move relatively and judge the size and speed
+ * of motion correctly.
+ */
+template <class RobotDriver>
+static GoalType generateSimpleGoal(RobotDriver* robot_driver,
```
hslusarek: In my opinion this function now violates the single responsibility principle. I suggest to add an additional function "on a higher abstraction level" which:
- at first ensures that the robot does not move, 
- generates the `alternating_position_shifts`, 
- fetches the `getMockRobotJointPositions()` and
- then calls the `generateSimpleGoal()` function with the `alternating_position_shifts[position_shift_index] + joint_positions[0]` parameter as argument.

pilz_control/test/pjtc_test_helper.h
```diff
@@ -155,6 +135,55 @@ static bool updateUntilRobotMotion(RobotDriver* robot_driver,
...
+                                   const ros::Duration& goal_duration = ros::Duration(DEFAULT_GOAL_DURATION_SEC),
+                                   const float size_multiplier = 1)
```
hslusarek: What is this argument to supposed to tell me?

pilz_control/test/pjtc_test_helper.h
```diff
@@ -155,6 +135,55 @@ static bool updateUntilRobotMotion(RobotDriver* robot_driver,
...
+  std::vector<double> joint_positions(NUM_JOINTS);
+  for (unsigned int i = 0; i < NUM_JOINTS; ++i)
+  {
+    joint_positions[i] = joint_data.at(i).pos;
+  }
+  return joint_positions;
+}
+
+/**
+ * @brief Cycle through four different postion deltas to always get a motion, even if last goal fails.
```
hslusarek: This is more a implementation detail and not a documentation of the function. Maybe add this sentence under a doxygen note tag?

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
martiniil: We only check acceleration. Inserted `std::abs` accordingly.

pilz_control/test/acceptance_test_acceleration_limit.py
martiniil: I don't know. Why does it matter?

pilz_control/test/acceptance_test_acceleration_limit.py
martiniil: Done.

pilz_control/test/acceptance_test_acceleration_limit.py
martiniil: Done.

pilz_control/test/acceptance_test_acceleration_limit.py
martiniil: Done.

pilz_control/test/acceptance_test_acceleration_limit.py
martiniil: Done.

pilz_control/test/acceptance_test_acceleration_limit.py
martiniil: outdated.

pilz_control/test/acceptance_test_acceleration_limit.py
martiniil: outdated.

pilz_control/test/acceptance_test_acceleration_limit.py
martiniil: outdated.

pilz_control/spec.dox
martiniil: We only limit acceleration.

pilz_control/test/acceptance_test_acceleration_limit.py
martiniil: Done.

pilz_control/test/acceptance_test_acceleration_limit.py
martiniil: outdated.

pilz_control/test/acceptance_test_acceleration_limit.py
martiniil: Yes. I think we can also consolidate parts used in both the acceptance- and the integrationtests. fyi @ct2034 @SansoneG

pilz_control/test/acceptance_test_acceleration_limit.py
martiniil: Done.

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
```diff
@@ -27,6 +27,50 @@ static constexpr double SPEED_LIMIT_NOT_ACTIVATED{ -1.0 };
...
+ *
+ * @param nh NodeHandle to access parameter server.
+ * @param joint_names Vector of Strings for all joint names to get the acceleration limits for.
+ * @return std::vector<double> Requested acceleration limits as vector. Has the same length as param 'joint_names'.
+ *
+ * @throw InvalidParameterException Requested values not found on param server.
+ */
+std::vector<double> getJointAccelerationLimits(const ros::NodeHandle& nh, const std::vector<std::string> joint_names)
+{
+  const std::string joint_limits_naming_prefix{ "/joint_limits/" };
```
martiniil: If I understand this correctly, the intention is to read the parameters loaded from `prbt_moveit_config`. This adds an `exec_depend` on `prbt_moveit_config` to `pilz_control`. I think this is wrong, because it introduces a circular dependency.

martiniil: Reminder: The controller limits should be higher than the planner limits.

ct2034: As agreed, we will use the limits defined in `prbt_support/config/manipulator_controller.yaml`. But we will not need a exec_depend on it, because this controller is as it is not standalone. It is only a plugin that will be loaded from `robot.launch` in `prbt_support`, anyway.

pilz_control/test/pjtc_test_helper.h
```diff
@@ -155,6 +135,55 @@ static bool updateUntilRobotMotion(RobotDriver* robot_driver,
...
+{
+  return waitFor([robot_driver]() { return !robot_driver->isRobotMoving(); }, movement_timeout,
+                 [robot_driver]() { robot_driver->update(); });
+}
+
+template <class RobotDriver>
+static std::vector<double> getMockRobotJointPositions(RobotDriver* robot_driver)
+{
+  robot_driver->update();
+  std::array<JointData, NUM_JOINTS> joint_data = robot_driver->getRobotMock().read();
```
hslusarek: Can be `const`.

pilz_control/test/robot_driver_mock.h
```diff
@@ -75,6 +77,12 @@ bool RobotDriverMock<ControllerManager>::isRobotMoving() const
   return robot_.isMoving();
 }
 
+template <class ControllerManager>
+RobotMock RobotDriverMock<ControllerManager>::getRobotMock()
```
hslusarek: Is it really the intention to return a _copy_ of the internally stored `RobotMock` or is the "real" intention to return a  const reference/pointer?

Also: The function can be inlined.

pilz_control/test/robot_mock.h
```diff
@@ -43,7 +43,7 @@ class RobotMock : public hardware_interface::RobotHW
 public:
   RobotMock();
 
-  void read();
+  std::array<JointData, NUM_JOINTS> read();
```
hslusarek: Function should be `const`.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
```diff
@@ -131,17 +151,55 @@ TEST_F(PilzJointTrajectoryControllerTest, testD0Destructor)
...
 {
-  ASSERT_TRUE(manager_->loadController()) << "Failed to initialize the controller.";
+  // test setup
+  ros::NodeHandle nh{ "~" };
+  std::vector<std::string> joint_names;
+  ASSERT_TRUE(nh.getParam(CONTROLLER_JOINT_NAMES_PARAM, joint_names));
+  ASSERT_FALSE(joint_names.empty());
+
+  // test with existing acc limits
+  std::vector<double> acceleration_limits = getJointAccelerationLimits(nh, joint_names);
```
hslusarek: Should be `const`.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
```diff
@@ -118,9 +120,27 @@ testing::AssertionResult PilzJointTrajectoryControllerTest::isControllerInUnhold
...
-//    The actual tests start here    //
-///////////////////////////////////////
+//////////////////////////////////
+//  Testing of Initialization   //
+//////////////////////////////////
+
+/**
+ * @brief Test if test environemt is initialized successfully.
+ */
+TEST_F(PilzJointTrajectoryControllerTest, testInitializiation)
```
hslusarek, line 131: Tests can be shuffled like described by the [gtest framework](https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#shuffling-the-tests), so what purpose does this test fulfill, if the tests are executed in random order?

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
```diff
@@ -131,17 +151,57 @@ TEST_F(PilzJointTrajectoryControllerTest, testD0Destructor)
...
+ *   Test if correct errors are thrown if parameters can not be correctly read.
+ * }
+ */
+TEST_F(PilzJointTrajectoryControllerTest, testGetJointAccelerationLimitsException)
+{
+  ros::NodeHandle nh{ "~" };
 
-  EXPECT_TRUE(action_client_.waitForActionServer());
+  // testing behaviour if acc limit can not be read
+  std::vector<std::string> joint_names_no_acc_limit = { "joint_with_undefined_max_acc" };
```
hslusarek: Should be `const`.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
```diff
@@ -131,17 +151,57 @@ TEST_F(PilzJointTrajectoryControllerTest, testD0Destructor)
...
+  EXPECT_THROW(getJointAccelerationLimits(nh, joint_names_no_acc_limit), ros::InvalidParameterException);
+
+  // testing behaviour if `has_acceleration_limits` is undefined
+  std::vector<std::string> joint_names_has_acc_lim_undefined = { "joint_with_undefined_has_acc_lim" };
```
hslusarek: Should be `const`.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
```diff
@@ -118,9 +120,27 @@ testing::AssertionResult PilzJointTrajectoryControllerTest::isControllerInUnhold
...
+  EXPECT_TRUE(ros::service::exists(CONTROLLER_NAMESPACE + HOLD_SERVICE, true));
+  EXPECT_TRUE(ros::service::exists(CONTROLLER_NAMESPACE + UNHOLD_SERVICE, true));
+  EXPECT_TRUE(ros::service::exists(CONTROLLER_NAMESPACE + IS_EXECUTING_SERVICE, true));
+
+  EXPECT_TRUE(action_client_.waitForActionServer());
+}
+
+/////////////////////////////////////////
+//    Testing of additional methods    //
+/////////////////////////////////////////
```
hslusarek: If we have a test file consisting of so many tests, that we need comments to visually "split" the file, maybe it is time to actually split the tests into separate test files. Currently the file consists of 654 lines...

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
```diff
@@ -131,17 +151,57 @@ TEST_F(PilzJointTrajectoryControllerTest, testD0Destructor)
...
 {
-  ASSERT_TRUE(manager_->loadController()) << "Failed to initialize the controller.";
+  // test setup
+  ros::NodeHandle nh{ "~" };
+  std::vector<std::string> joint_names;
+  ASSERT_TRUE(nh.getParam(CONTROLLER_JOINT_NAMES_PARAM, joint_names));
+  ASSERT_FALSE(joint_names.empty());
+
+  // test with existing acc limits
+  std::vector<boost::optional<double>> acceleration_limits = getJointAccelerationLimits(nh, joint_names);
```
hslusarek: Can be `const`

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
```diff
@@ -131,17 +151,57 @@ TEST_F(PilzJointTrajectoryControllerTest, testD0Destructor)
...
+  EXPECT_EQ(joint_names.size(), acceleration_limits.size());
+  EXPECT_TRUE(acceleration_limits.at(0));
+  EXPECT_TRUE(acceleration_limits.at(1));
+  EXPECT_DOUBLE_EQ(*acceleration_limits.at(0), 3.49);
+  EXPECT_DOUBLE_EQ(*acceleration_limits.at(1), 3.49);  // as of pilz_control/test/config/joint_limits.yaml
+
+  // testing behaviour if `has_acceleration_limits` is false
+  std::vector<std::string> joint_names_has_acc_lim_false = { "joint_with_has_acc_lim_false" };
```
hslusarek: Should be `const`.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
```diff
@@ -131,17 +151,57 @@ TEST_F(PilzJointTrajectoryControllerTest, testD0Destructor)
...
+
+  // test with existing acc limits
+  std::vector<boost::optional<double>> acceleration_limits = getJointAccelerationLimits(nh, joint_names);
+  EXPECT_EQ(joint_names.size(), acceleration_limits.size());
+  EXPECT_TRUE(acceleration_limits.at(0));
+  EXPECT_TRUE(acceleration_limits.at(1));
+  EXPECT_DOUBLE_EQ(*acceleration_limits.at(0), 3.49);
+  EXPECT_DOUBLE_EQ(*acceleration_limits.at(1), 3.49);  // as of pilz_control/test/config/joint_limits.yaml
+
+  // testing behaviour if `has_acceleration_limits` is false
```
hslusarek: It seems to me that you test two objectives in one test. Maybe split into two tests?

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
```diff
@@ -131,17 +151,57 @@ TEST_F(PilzJointTrajectoryControllerTest, testD0Destructor)
...
+  std::vector<std::string> joint_names_has_acc_lim_false = { "joint_with_has_acc_lim_false" };
+  std::vector<boost::optional<double>> acceleration_limits_has_acc_lim_false =
```
hslusarek: Should be `const`.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
```diff
@@ -355,6 +419,40 @@ TEST_F(PilzJointTrajectoryControllerTest, testGoalCancellingDuringHold)
...
+  std_srvs::TriggerRequest req;
+  std_srvs::TriggerResponse resp;
+  EXPECT_TRUE(manager_->triggerUnHold(req, resp));
+  EXPECT_TRUE(resp.success);
+  EXPECT_TRUE(isControllerInUnholdMode());
+
+  // Make sure, robot moves for slow motion
+  GoalType goal{ generateSimpleGoal<RobotDriver>(&robot_driver_, ros::Duration(DEFAULT_GOAL_DURATION_SEC)) };
+  action_client_.sendGoal(goal);
+  EXPECT_TRUE(manager_->controller_->is_executing());
```
hslusarek: Just because a line before you send the goal, the robot not necessarily executes already. Therefore, I suggest that you use the function `updateUntilRobotMotion()`.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
```diff
@@ -355,6 +419,40 @@ TEST_F(PilzJointTrajectoryControllerTest, testGoalCancellingDuringHold)
...
+  action_client_.waitForActionResult();
+  EXPECT_TRUE(updateUntilNoRobotMotion<RobotDriver>(&robot_driver_));
+
+  // Now sending a quicker motion which should trigger the acceleration limit and not move the robot
+  goal = generateSimpleGoal<RobotDriver>(&robot_driver_, ros::Duration(DEFAULT_GOAL_DURATION_SEC), 1E4);
+  EXPECT_TRUE(updateUntilNoRobotMotion<RobotDriver>(&robot_driver_));
```
hslusarek: You already checked this two lines above, so why do you check it again?

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
```diff
@@ -355,6 +419,40 @@ TEST_F(PilzJointTrajectoryControllerTest, testGoalCancellingDuringHold)
...
+  action_client_.sendGoal(goal);
+  EXPECT_FALSE(manager_->controller_->is_executing());
```
hslusarek: Just because a line before you send the goal, the robot not necessarily executes already. Therefore, I suggest that you use the function `updateUntilRobotMotion()`.

prbt_hardware_support/test/unit_tests/unittest_brake_test_utils.cpp
```diff
@@ -167,6 +167,13 @@ TEST(BrakeTestUtilsTest, testGetCurrentJointStates)
 
   auto expected_msg = joint_states_pub.getNextMessage();
 
+  // Move away from zero
+  joint_states_pub.setVelocity(0.1);
+  ros::Duration(0.1).sleep();
```
hslusarek: Can we get rid of the `sleep` by using some kind of barrier instead?

hslusarek (changes requested):
Review for unit- and integrationtests.

pilz_control/test/acceptance_test_acceleration_limit.py
martiniil: Done

prbt_hardware_support/test/unit_tests/unittest_brake_test_utils.cpp
martiniil: Sorry, this has to be a mistake from resolving a merge-conflict. I revert it.

ct2034:
@martiniil can you please fix `testintegrationtest_pilz_joint_trajectory_controller`. I think it worked for me before your latest changes. Hope its not my fault 😬

pilz_control/test/acceptance_test_acceleration_limit.py
SansoneG: resolved

pilz_control/test/config/joint_limits.yaml
martiniil: Added `max_acceleration` entry

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
martiniil: Done.

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
martiniil: Done

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
martiniil: Done.

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
martiniil: Done.

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
martiniil: Done.

pilz_control/test/pjtc_test_helper.h
martiniil: Done.

pilz_control/test/pjtc_test_helper.h
martiniil: outdated.

pilz_control/test/robot_driver_mock.h
martiniil: This internally stored `RobotMock` should not be accessible. I replaced this function by `getJointPositions()`.

pilz_control/test/robot_mock.h
martiniil: Done.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
martiniil: Done.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
martiniil: Done.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
martiniil: Done.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
martiniil: Done.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
martiniil: Done.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
martiniil: Done.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
martiniil: The comment is wrong. The `init()` method of the controller is tested here. I remove the comment.

pilz_control/test/pjtc_test_helper.h
martiniil: DOne.

pilz_control/test/pjtc_test_helper.h
martiniil: Renamed into `distance_scaling_factor`.

pilz_control/test/pjtc_test_helper.h
martiniil: Done.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
martiniil: Done.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
martiniil: Not necessary to check `is_executing` here.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
martiniil: Not necessary/useful to check `is_executing` here.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
martiniil: Done.

pilz_control/test/unittest_pilz_joint_trajectory_controller.cpp
martiniil: I introduced 2 new files in order to reduced the tests per file.

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
martiniil: Sounds good. The values we process here are generated in the canopen driver / hardware_interface. Would it make sense to start there with a pull request? Let's continue discussing about that.

pilz_control/include/pilz_control/pilz_joint_trajectory_controller_impl.h
```diff
@@ -53,10 +53,14 @@ static const std::string MONITOR_CARTESIAN_SPEED_SERVICE_NAME{ "monitor_cartesia
...
 
+/**
+ * @brief Calculate absolute acceleration based on two velocities (`current_velocity` and `old_velocity`)
+ *  and the time interval between them (`delta_t`)
+ */
 inline double calculateAcceleration(const double& current_velocity, const double& old_velocity,
                                     const ros::Duration& delta_t)
 {
-  return (std::abs(current_velocity) - std::abs(old_velocity)) / delta_t.toSec();
+  return (std::abs(current_velocity - old_velocity)) / delta_t.toSec();
```
martiniil: @ct2034 Do you really want to limit also the deceleration? If yes, why?

martiniil:
Continue here: https://github.com/PilzDE/pilz_robots/pull/414

PR #408, opened 2020-06-10
Feature/extend robot mock

martiniil:
This change allows to perform a Python "acceptance-test" involving the `pilz_joint_trajectory_controller` with the `robot_mock` - via `robot_mock.launch`. Main changes:
- Launch `robot_mock` node in `robot_mock.launch`
- Add `joint_state_controller`
- Remove end-effector link from `robot_mock_monitored_links.yaml`

Both acceptance-tests in pilz_control can now be run with the robot_mock in order to pre-validate them without real robot hardware. E.g.
```
roslaunch pilz_control robot_mock.launch
rosrun pilz_control acceptance_test_acceleration_limit.py
```
or
```
roslaunch pilz_control robot_mock.launch
roslaunch pilz_control joint_states_speed_observer.launch robot_name:=robot_mock
rosrun pilz_control acceptance_test_speed_monitoring.py
```


pilz_control/test/unittest_joint_states_speed_observer.test
```diff
@@ -23,9 +23,9 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
   <param name="robot_description" textfile="$(find pilz_control)/test/urdf/robot_mock.urdf"/>
   <param name="robot_description_semantic" textfile="$(find pilz_control)/test/urdf/robot_mock.srdf"/>
 
-  <rosparam ns="prbt" command="load" file="$(find pilz_control)/test/config/robot_mock_monitored_links.yaml" />
-
-  <node pkg="pilz_control" type="joint_states_speed_observer" name="joint_states_speed_observer" output="screen"/>
+  <include file="$(find pilz_control)/test_utils/launch/joint_states_speed_observer.launch">
+      <arg name="robot_name" value="robot_mock" />
+  </include>
```
hslusarek, line 28: This unittest currently relies on the `joint_state_publisher_mock.cpp` and does not work because it needs to set the correct joint velocity see [`joint_state_pub_.setJoint1Velocity(angular_velocity)`](https://github.com/PilzDE/pilz_robots/blob/cb782c272ee68141c23b0a8bc03f0a0eb14d3807/pilz_control/test/unittest_joint_states_speed_observer.cpp#L131)

Furthermore, you wrote:
> This change allows to perform a Python "acceptance-test" involving the pilz_joint_trajectory_controller with the robot_mock - via robot_mock.launch.

So, why are you changing this unittest and not an acceptance test?

hslusarek (changes requested):
Please, provide an acceptance test which actually needs your changes, so that it easier to evaluate if your changes are correct and needed. 
Furthermore, if you want to change the `unittest_joint_states_speed_observer` please make sure that the test runs. The test currently needs the [` pilz_testutils::JointStatePublisherMock`](https://github.com/PilzDE/pilz_robots/blob/946c57cf57ed76dd1080d2ed9c7e4f3631358f6e/pilz_control/test/unittest_joint_states_speed_observer.cpp#L101). In order, to make the test run, you have to remove this dependency.

pilz_control/test/unittest_joint_states_speed_observer.test
martiniil: I did only refactor the test launch-file. It should not change the unittest. Let's wait for the new travis builds. Then we will see, if there is a problem.

martiniil:
@hslusarek  Thanks for your review. This change is not really needed by any test. It makes writing acceptance-tests easier. See the updated PR description.

pilz_control/test/acceptance_test_speed_monitoring.py
```diff
@@ -78,17 +79,20 @@ class AcceptancetestSpeedMonitoring(unittest.TestCase):
     """
 
     def setUp(self):
-        param_name = CONTROLLER_NS + _JOINT_NAMES_PARAMETER
+        controller_ns = rospy.get_param(_CONTROLLER_NS_PARAMETER, _DEFAULT_CONTROLLER_NS)
```
hslusarek, line 82: You stated in the PR description that it is possible to execute the acceptance tests without real hardware. I tried the `acceptance_test_speed_monitoring.py` like you described. In other words, I executed:
```
roslaunch pilz_control robot_mock.launch
rosrun pilz_control acceptance_test_speed_monitoring.py
```
I got an error. For details see attached picture:
error_acceptance_test_speed_monitoring py

I'm guessing the test is not intended to be used without `joint_states_speed_observer`? In other words, I always have to launch the `joint_states_speed_observer`:
```
roslaunch pilz_control joint_states_speed_observer.launch robot_name:=robot_mock
```

pilz_control/test/acceptance_test_acceleration_limit.py
```diff
@@ -24,10 +24,11 @@
 from holding_mode_service_wrapper import HoldingModeServiceWrapper
 from trajectory_dispatcher import TrajectoryDispatcher
 
-# Change the following two lines if you run a different robot
-CONTROLLER_NS = '/prbt'
-CONTROLLER_NAME = 'manipulator_joint_trajectory_controller'
+DEFAULT_CONTROLLER_NS = '/prbt'
```
hslusarek, line 27: I very much like the idea that we now are able to execute these tests without real hardware. I, therefore, suggest that we add a comment in the `acceptance_test_acceleration_limit.md` and `acceptance_test_speed_monitoring.md` on how to start the tests without real hardware. In other words, I suggest to add the things that you already stated in your PR description. What do you think? 

We could also include these tests in the CI? What do you think about that?

hslusarek (changes requested):
Code review.
Thx for all your work. I really like the changes you made. I have two minor suggestions, see the review comments.
You can decide for yourself if you want to incorporate my suggested changes or not. If you are done, you can simply merge.

pilz_control/test/acceptance_test_speed_monitoring.py
martiniil: Yes. I added a description to the .md-file.

pilz_control/test/acceptance_test_acceleration_limit.py
martiniil: Thanks. I added them. (edit: the description)

I think, before we can add these test to the CI, we have to pay special attention if they can run reliably on a build server. But it is a good point. This could be a new maintenance task?

hslusarek: Sounds good to me.

hslusarek (approved):

PR #410, opened 2020-06-15
Improving error response in case of IP adress failures.

### LINKED ISSUE #382: Failing Modbus connection often produces no error message
### Commit
b5d88cf6b3d0628ec25a5720522ed4a7d213c4fe

### Steps to reproduce
1. make sure Modbus connection can be established.
2. disconnect the ethernet cable
3. try to establish Modbus connection again

### Other
The same behaviour happens when:
* wrong port is set
* ethernet disconnected
* if own IP is in the wrong subnet
(Only if the target IP is wrong but in the correct subnet and the own IP is in the correct subnet, a meaningful error message is produced)

### Expected behaviour
An error message is shown that describes the problem directly

### Observed behaviour
No error is produced. In fact, the console looks exactly the same as when a connection is established successfully.

saheinisch:
Changes in Constructor of class LibModbusClient in libmodbus_client.cpp. 
Before a modbus connection is established a dummy TCP connection is made with a zero timeout. This is repeated 15 times every 100ms until success. If connection attempt was successful the connection is closed. If it was not successful the constructor returns false.

### DIFF
--- a/prbt_hardware_support/include/prbt_hardware_support/libmodbus_client.h
+++ b/prbt_hardware_support/include/prbt_hardware_support/libmodbus_client.h
@@ -72,7 +72,7 @@ private:
  * @return true if the connection to the server succeeded
  * @return false if the connection to the server failed
  */
-bool checkIPConnection(const char* ip, const unsigned int port, double timeout, const unsigned int n_retries);
+bool checkIPConnection(const char* ip, const unsigned int port);
 
 }  // namespace prbt_hardware_support
 
--- a/prbt_hardware_support/include/prbt_hardware_support/libmodbus_client.h
+++ b/prbt_hardware_support/include/prbt_hardware_support/libmodbus_client.h
@@ -60,6 +60,7 @@ public:
 
 private:
   modbus_t* modbus_connection_{ nullptr };
+  bool checkIPConnection(const char* ip, unsigned int port, unsigned int timeout, unsigned int retries);
 };
 
 }  // namespace prbt_hardware_support
--- a/prbt_hardware_support/include/prbt_hardware_support/libmodbus_client.h
+++ b/prbt_hardware_support/include/prbt_hardware_support/libmodbus_client.h
@@ -60,7 +60,7 @@ public:
 
 private:
   modbus_t* modbus_connection_{ nullptr };
-  bool checkIPConnection(const char* ip, unsigned int port, unsigned int timeout, unsigned int retries);
+  bool checkIPConnection(const char* ip, unsigned int port, unsigned int timeout, unsigned int retries) const;
 };
 
 }  // namespace prbt_hardware_support
--- a/prbt_hardware_support/include/prbt_hardware_support/libmodbus_client.h
+++ b/prbt_hardware_support/include/prbt_hardware_support/libmodbus_client.h
@@ -63,15 +63,15 @@ private:
 };
 
 /**
-   * @brief Test the ip connection by connecting to the modbus server
-   *
-   * @param ip of the modbus server
-   * @param port of the modbus server
-   * @param timeout of the modbus connection attempt
-   * @param retries of the connection attempt lasting as long as defined in timout 
-   * @return true if the connection to the server succeeded
-   * @return false if the connection to the server failed
-   */
+ * @brief Test the ip connection by connecting to the modbus server
+ *
+ * @param ip of the modbus server
+ * @param port of the modbus server
+ * @param timeout of the modbus connection attempt
+ * @param retries of the connection attempt lasting as long as defined in timout
+ * @return true if the connection to the server succeeded
+ * @return false if the connection to the server failed
+ */
 bool checkIPConnection(const char* ip, unsigned int port, unsigned int timeout, unsigned int retries);
 
 }  // namespace prbt_hardware_support
--- a/prbt_hardware_support/include/prbt_hardware_support/libmodbus_client.h
+++ b/prbt_hardware_support/include/prbt_hardware_support/libmodbus_client.h
@@ -67,8 +67,6 @@ private:
  *
  * @param ip of the modbus server
  * @param port of the modbus server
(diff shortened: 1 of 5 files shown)


prbt_hardware_support/src/libmodbus_client.cpp
```diff
@@ -35,15 +36,45 @@ LibModbusClient::~LibModbusClient()
...
+    if (0  == retries)
+    {
+       ROS_ERROR_STREAM_NAMED("LibModbusClient","Could not establish modbus connection. Cable connected?");
+       return false;
+    }
+    usleep(100000); // wait 1 sec
+  }
+  ::close(sockfd); 
+  sleep(1);                                  //wait one second until port is released
+
```
alex-gee: I suggest moving this into a separate function, parameters could be the timeout and number of retries.

prbt_hardware_support/src/libmodbus_client.cpp
```diff
@@ -35,15 +36,45 @@ LibModbusClient::~LibModbusClient()
...
+  //try 20 times to connect within 100 mseconds
+  while ((connect(sockfd,(const sockaddr *) &serv_addr,sizeof(serv_addr)) < 0) && ( 0 != retries))
+  {
+    retries--;    
+    if (0  == retries)
+    {
+       ROS_ERROR_STREAM_NAMED("LibModbusClient","Could not establish modbus connection. Cable connected?");
+       return false;
+    }
+    usleep(100000); // wait 1 sec
```
alex-gee: Here usleep, later sleep. Why both?

saheinisch: usleep to obtain connection attempts every 100ms. 
Sleep cause after closing the socket the host needs some time until the socket can be reused. I have taken 1 second , which is an eternity for todays machine, to be shure the following connection will work. On the other hand one additional second for establishing the connection is acceptable for the user on startup. So no need here to use usleep(1000000) instead of sleep(1)

prbt_hardware_support/src/libmodbus_client.cpp
saheinisch: Ok this is a good idea.

prbt_hardware_support/src/libmodbus_client.cpp
```diff
@@ -35,15 +36,45 @@ LibModbusClient::~LibModbusClient()
...
+  struct sockaddr_in serv_addr;
+  
+  sockfd = socket(AF_INET, SOCK_STREAM, 0);      // Create socket for testing purpose
+  bzero((char *) &serv_addr, sizeof(serv_addr)); 
+  serv_addr.sin_family = AF_INET;                   
+  serv_addr.sin_port = htons((short unsigned int)port);
+  
+  arg = fcntl(sockfd, F_GETFL, NULL); 
+  arg |= O_NONBLOCK; 
+  fcntl(sockfd, F_SETFL, arg);                   //set connection to non blocking, no timeout
```
cdoehn: Maybe better rename the auxiliary variable arg to result

```suggestion
  long   result; // auxiliary variable
  int    sockfd;
  int    retries;
  struct sockaddr_in serv_addr;
  
  sockfd = socket(AF_INET, SOCK_STREAM, 0);      // Create socket for testing purpose
  bzero((char *) &serv_addr, sizeof(serv_addr)); 
  serv_addr.sin_family = AF_INET;                   
  serv_addr.sin_port = htons((short unsigned int)port);
  
  result = fcntl(sockfd, F_GETFL, NULL); 
  result |= O_NONBLOCK; 
  fcntl(sockfd, F_SETFL, result);                   //set connection to non blocking, no timeout
```

prbt_hardware_support/src/libmodbus_client.cpp
cdoehn:
```suggestion
    usleep(100000); // wait 100 ms
```

cdoehn (dismissed):
Please fix clang format issues and look at my suggested changes

cdoehn:
Fix clang format issues

saheinisch (commented):
Would like to not change it until there is a technical reason.

saheinisch:
> Please fix clang format issues and look at my suggested changes

My codium does not display any clang format issue.  
Where are those format issues occurring?

agutenkunst:
> Would like to not change it until there is a technical reason. sleep usleep

This refers to what?

agutenkunst:
> > Please fix clang format issues and look at my suggested changes
> 
> My codium does not display any clang format issue.
> Where are those format issues occurring?

On Travis I suppose.

prbt_hardware_support/include/prbt_hardware_support/libmodbus_client.h
```diff
@@ -60,6 +60,7 @@ class LibModbusClient : public ModbusClient
 
 private:
   modbus_t* modbus_connection_{ nullptr };
+  bool checkIPConnection(const char* ip, unsigned int port, unsigned int timeout, unsigned int retries);
```
agutenkunst: Does not need to be member. But if it could be static and const.

saheinisch: What do you mean exactly?

agutenkunst: As far as I understand this function does not use any member variables and thus does not need to be a member function.

saheinisch: Thank you. Ok, I made it constant. Its pushed already.

agutenkunst:
I suggest using calling `checkIPConnection` somewhere from productive code. :stuck_out_tongue_winking_eye:

agutenkunst (approved):
Test may not be optimal but due to the fact that this is just a workaround since it will be obsolete with Ubuntu 20 and its libmodbus 3.1.6 where the modbus timeout can be set directly this is acceptable.

PR #413, opened 2020-06-19
Add infinite retries on modbus connect and set it as default.

agutenkunst:
As discussed with @jschleicher this is probably a better to handle a connection error.

@hslusarek @martiniil 
Before this PR the modbus node exists after the retries and shuts down the ROS system due to `required=true`. Do you guys have any additional considerations?


prbt_hardware_support/src/pilz_modbus_client.cpp
```diff
@@ -38,21 +38,22 @@ PilzModbusClient::PilzModbusClient(ros::NodeHandle& nh, const std::vector<unsign
...
   {
+    ++retry_n;
+
     if (init(ip, port))
     {
       return true;
     }
 
-    ROS_ERROR_STREAM("Connection to " << ip << ":" << port << " failed. Try(" << retry_n + 1 << "/" << retries << ")");
+    ROS_WARN_STREAM_COND(retries == -1, "Connection to " << ip << ":" << port << " failed.");
```
jschleicher: Add a hint here how to solve in the warning message (and check the output - I cannot remember this message during our tests at the real robot?):
https://github.com/PilzDE/pilz_robots/blob/a2ea32b49d1d431100604616fa8aa129916bc2a7/prbt_hardware_support/src/pilz_modbus_client_node.cpp#L127-L128

jschleicher (commented):
Looks good. Thank You! @saheinisch Please take a look, if this fixes the ROS shutdown.

martiniil:
Looks good. @jschleicher should we wait for this with the release?

prbt_hardware_support/src/pilz_modbus_client.cpp
saheinisch: Looks good, lets append it to the messages

saheinisch (approved):

jschleicher (approved):

PR #416, opened 2020-06-22
Add is_executing service to architecture diagram

martiniil:


jschleicher (commented):
Should the new speed observer be also part of this diagram?
@SansoneG

martiniil:
> Should the new speed observer be also part of this diagram?
> @SansoneG

Just to clarify: The `joint_state_speed_observer` is only part of test-code.

ct2034 (approved):
> Should the new speed observer be also part of this diagram?
> @SansoneG

As far as I understand the latest architecture, nothing has changed for the speed observer. The observation of speeds is done within the controller and has not added any new communication interfaces that would need to be added to this. The (old) speed observer node is only used for testing.

PR #422, opened 2020-06-24
Add missing header to pilz_testutils

agutenkunst:
Should fix 
http://build.ros.org/job/Mbin_uB64__pilz_testutils__ubuntu_bionic_amd64__binary/22/

Release needed after this.


martiniil (approved):
TY!

PR #411, opened 2020-06-16
Refactor invokeIsExecuting() function

hslusarek:
The unittest `unittest_pilz_joint_trajectory_controller.cpp ` consists of unnecessary complex and complicated code regarding the `invokeIsExecuting()` function which leads to code bloat in the tests itself. This PR tries to remedy the code bloat.

PR #424, opened 2020-06-29
Refactor invokeIsExecuting() function

hslusarek:
The `unittest_pilz_joint_trajectory_controller_is_executing.cpp ` consists of unnecessary complex and complicated code regarding the `invokeIsExecuting()` function which leads to code bloat in the tests itself. To improve the readability and understandability of the tests, this PR tries to remedy the code bloat.


hslusarek:
This PR continuous PR #411, which is now outdated, due to changes on the melodic-devel branch.

martiniil (changes requested):
I like that this reduces complexity of the test.

However I would avoid throwing exceptions in a test. The outcome of this would be:
```
[Testcase: testunittest_pilz_joint_trajectory_controller_is_executing] ... ERROR!
ERROR: max time [60.0s] allotted for test [unittest_pilz_joint_trajectory_controller_is_executing] of type [pilz_control/unittest_pilz_joint_trajectory_controller_is_executing]
```
And I still think it is a valid expectation, that the callback returns true.

Suggestion: We could drop checking the return value here, with the argument that this is tested on the integration-level, where the callback is invoked via service calls. What do you think?

Issue #415, opened 2020-06-19
fatal error: 'gmock/gmock.h' file not found


v4hn:
### Commit

`melodic-devel`

### Steps to reproduce

1. Probably only by setting up a non-debian Linux
2. Build

### Observed behavior

```
pilz_robots/pilz_testutils/src/joint_state_publisher_mock.cpp:23:10: fatal error: 'gmock/gmock.h' file not found
```

Looking at the respective `CMakeLists.txt`, `GMOCK_INCLUDE_DIRS` is not mentioned anywhere.
In another project I had to switch from `catkin_add_gtest` to `catkin_add_gmock` to fix this, but as this is a library, that wouldn't work. You probably have to add gmock manually.

I'm guessing Ubuntu/Debian might provide a symlink `/usr/include/gmock`?

**Addendum**:

Adding

```
target_include_directories(${PROJECT_NAME} PRIVATE "${GMOCK_INCLUDE_DIRS}" "${GTEST_INCLUDE_DIRS}")
```

fixes the problem on my side, but of course you would need to make sure that the variables are actually set (probably only if `CATKIN_ENABLE_TESTING` is set?)

hslusarek:
Thanks for your input. I looked into `joint_state_publisher_mock.cpp` and discovered that the `#include <gmock/gmock.h>` is actually not needed. I opened a PR #425 to remove the include.

hslusarek:
Ok, I also opened PR #426 which should add the missing includes, etc.

@v4hn Maybe you can try it out and see if it works for you.

v4hn:
Sorry, but I'm confused. Why do you propose to add the dependencies when the include is not necessary?

hslusarek:
> Sorry, but I'm confused. Why do you propose to add the dependencies when the include is not necessary?

Sorry, for the confusing communication. The `joint_state_publisher_mock.cpp` file does _not_ need the `gmock` include. However, the `async_test` does need the `gmock` include, although it's missing the corresponding `gmock` include right now. 
For more information see my comment on PR https://github.com/PilzDE/pilz_robots/pull/426#issuecomment-651722245

v4hn:
> The `async_test` does need the `gmock` include because it uses `::testing::InvokeWithoutArgs`

Syntactically it does not need the include because it only defines preprocessor macros with the offending symbol, it does not use them.
As far as I can see, the whole library does not use a single gtest or gmock symbol in preprocessed code.

[...]

PR #425, opened 2020-06-29
Remove unnecessary include in joint_state_publisher_mock.cpp

hslusarek:
Triggered by issue #415, I looked at the `joint_state_publisher_mock.cpp` and couldn't find a reason why the `#include <gmock/gmock.h>` exists. To solve issue #415, I suggest to simply remove the include and all other unnecessary includes.


jschleicher (approved):
LGTM

hslusarek:
@jschleicher I now also added a missing gmock include in the `async_test.h` file. Have a look again, please. Thx.

PR #426, opened 2020-06-29
Add gmock and gtest dependencies to library target

hslusarek:
Currently, the library target in [pilz_testutils/CMakeLists.txt](https://github.com/PilzDE/pilz_robots/blob/d7fa0f1a9244c7380203c187b311960d2edc9214/pilz_testutils/CMakeLists.txt#L54) is missing the gtest and gmock dependencies which causes build errors at some machines (see issue #415).
This PR tries to remedy the problems caused by the missing dependencies. 

**Note:**
The PR is based on the catkin cmake target [_catkin_add_executable_with_google_test()](https://github.com/ros/catkin/blob/afb0e2a2ae8c0c34c17a2df7b8e1219792246bf1/cmake/test/gtest.cmake#L154) from ROS.
Maybe in the future we can open a PR adding the newly introduced cmake function  `_catkin_add_library_gmock()` to the  catkin repo.

### DIFF
--- a/pilz_testutils/CMakeLists.txt
+++ b/pilz_testutils/CMakeLists.txt
@@ -39,6 +39,38 @@ if(CATKIN_ENABLE_CLANG_TIDY)
   endif()
 endif()
 
+##########################
+## Function definition ##
+#########################
+
+function(_catkin_add_library_gmock target)
+  if(NOT $GMOCK_FOUND AND NOT GMOCK_FROM_SOURCE_FOUND)
+    message(WARNING "skipping GMOCK '${target}' in project '${PROJECT_NAME}' because GMOCK was not found")
+    return()
+  endif()
+
+  if(NOT DEFINED CMAKE_RUNTIME_OUTPUT_DIRECTORY)
+    message(FATAL_ERROR "_catkin_add_library_gmock() must be called after catkin_package() so that default output directories for the libraries are defined")
+  endif()
+
+  cmake_parse_arguments(ARG "EXCLUDE_FROM_ALL" "" "" ${ARGN})
+
+  list(APPEND GMOCK_INCLUDE_DIRS ${GTEST_INCLUDE_DIRS})
+  list(APPEND GMOCK_LIBRARY_DIRS ${GTEST_LIBRARY_DIRS})
+  list(APPEND GMOCK_LIBRARIES ${GTEST_LIBRARIES})
+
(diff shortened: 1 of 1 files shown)


v4hn:
As you stated in #425 gmock is not needed here, so why add the whole gmock stuff?

Strictly speaking [the only gtest include](https://github.com/PilzDE/pilz_robots/blob/melodic-devel/pilz_testutils/include/pilz_testutils/async_test.h#L23) is also not necessary. The header only defines *preprocessor macros* using gtest symbols. So if you can add the requirement that all code using this header must also include gtest, you can drop the whole pull-request here.

hslusarek:
> Strictly speaking [the only gtest include](https://github.com/PilzDE/pilz_robots/blob/melodic-devel/pilz_testutils/include/pilz_testutils/async_test.h#L23) is also not necessary.

The `async_test.h` actually needs a `gmock` and not a `gtest` include because [`::testing::InvokeWithoutArgs`], used by `async_test.h`, is part of the gmock framework (seen github.com/googletest](https://github.com/google/googletest/blob/master/googlemock/include/gmock/gmock-actions.h#L1261)).

I suggest that we separate the macros needing the `gmock` dependencies into a separate header. This way, the `pilz_testutils` lib does not need the `gmock` include. Furthermore, the `gmock` dependency can be clearly stated in the header containing the macros, and like @v4hn stated we can then drop this PR.
@agutenkunst What do you think?

jschleicher:
I like the idea of @v4hn to drop the include and add a note to our [documentation](http://docs.ros.org/melodic/api/pilz_testutils/html/pilz_testutils/classtesting_1_1AsyncTest.html) that the user of `pilz_testutils` needs to depend/include gmock, see for example https://github.com/PilzDE/pilz_robots/blob/cb782c272ee68141c23b0a8bc03f0a0eb14d3807/pilz_control/CMakeLists.txt#L104

agutenkunst:
Would 
`<build_export_depend>` be useful here?

hslusarek:
I discussed the matter with @agutenkunst and we decided to favor a compromise:
Async_test is going to be header only (see PR #430). This removes `async_test` from the `pilz_testutils` lib build and at the same time we can clearly state the dependency to gmock in the `async_test.h` file via `#include <gmock/gmock.h>`.
This PR can be closed as a consequence. Thx to everyone for the participation in the discussion.

PR #428, opened 2020-06-30
Remove unneeded includes from pilz_testutils

agutenkunst:
See https://github.com/PilzDE/pilz_robots/issues/415


hslusarek:
See https://github.com/PilzDE/pilz_robots/pull/426#issuecomment-652344410

PR #430, opened 2020-07-01
Make AsyncTest header only

hslusarek:
The `async_test` depends on gmock when it comes to its MACROS. This dependency has to be stated clearly using the corresponding `#include <gmock/gmock.h>` in the `async_test.h` file. If the dependency is not stated in `async_test.h` an unexperienced user might encounter not understandable compiler errors when using the `async_test.h` and forgetting to add the `#include <gmock/gmock.h>` to the test file. 
However, when building the `pilz_testutils` lib, it is not necessary to already include all build dependencies in the build because this is automatically done by the tests using the `async_test` via the CMAKE functions `catkin_add_gmock()` or `add_rostest_gmock()`. Furthermore, it makes no sense to build all testutils into one big library. For the mentioned reasoned it was decided to make the `async_test` to a header only file and, therefore, effectively remove it from the `pilz_testutils` lib build.

**Note:**
This PR is a consequence of issue #415 and the discussions in PR #426.


agutenkunst (changes requested):
Please:

- [x] Update changelog
- [ ] Open a PR on https://github.com/PilzDE/pilz_industrial_motion/blob/melodic-devel/pilz_trajectory_generation/test/integrationtest_sequence_action_capability.cpp

agutenkunst (approved):

PR #436, opened 2020-07-02
Add tolerance to speed-limit acceptance-test

martiniil:
This should compensate for noise in the movement of the real robot. So far the speed monitor allows all movements up to `0.25m/s`, which is way too sharp. We still need to quantify how much this should be lower than the actual `0.25m/s`.


jschleicher (approved):
LGTM

PR #447, opened 2020-07-29
Update acceptancetest_stop1.md

martiniil:
- We need to press acknowledge again after startup, else the controller is in holding mode.
- Other minor fixes


jschleicher:
>     * We need to press acknowledge again after startup, else the controller is in holding mode.

Yes, that's the current behaviour - just tested with the PRBT. But from a user's perspective an incorrect assumption. Why shouldn't I be able to move the robot, if acknowledge is pressed during launch? 
I don't like to change the acceptance test but rather fix the implementation. Is this a regression?

jschleicher (approved):
Approving except second acknowledge, merging.

martiniil:
> 
> 
> > ```
> > * We need to press acknowledge again after startup, else the controller is in holding mode.
> > ```
> 
> Yes, that's the current behaviour - just tested with the PRBT. But from a user's perspective an incorrect assumption. Why shouldn't I be able to move the robot, if acknowledge is pressed during launch?
> I don't like to change the acceptance test but rather fix the implementation. Is this a regression?

We talked about this in the past (I think with @dbakovic). It is not a regression. But it doesn't matter since we agree that it should be fixed.

PR #465, opened 2020-10-26
Fix NiceMock and StrickMock issue in LoggerMock

hslusarek:
This PR fixes the problem that the different mock types like `StricktMock`, `NiceMock`, etc. cannot be properly used in combination with the `LoggerMock`.

### DIFF
--- a/pilz_testutils/include/pilz_testutils/logger_mock.h
+++ b/pilz_testutils/include/pilz_testutils/logger_mock.h
@@ -20,10 +20,15 @@
 
 #include <log4cxx/logger.h>
 
+#include <gmock/gmock.h>
+
 #include <pilz_testutils/mock_appender.h>
 
 namespace pilz_testutils
 {
+using std::placeholders::_1;
+using std::placeholders::_2;
+
 /**
  * @brief Class to be used in tests with logging checks.
  *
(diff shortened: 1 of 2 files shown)


pilz_testutils/include/pilz_testutils/logger_mock.h
```diff
@@ -53,11 +58,13 @@ namespace pilz_testutils
  *
  * const std::string LOG_MSG_RECEIVED_EVENT{ "logger_called_event" };
  *
- * pilz_testutils::LoggerMock ros_log_mock;
+ * StrictMock<pilz_testutils::LoggerMock> ros_log_mock;
  *
- * EXPECT_LOG(*mock_logger, WARN, "Your warning text")
+ * EXPECT_WARN_LOG(mock_logger, "Your warning text")
```
martiniil, line 63:
```suggestion
 * EXPECT_WARN_LOG(ros_log_mock, "Your warning text")
```
correct?

pilz_testutils/include/pilz_testutils/logger_mock.h
```diff
@@ -53,11 +58,13 @@ namespace pilz_testutils
...
  *            .WillOnce(ACTION_OPEN_BARRIER_VOID(LOG_MSG_RECEIVED_EVENT));
  *
+ * EXPECT_ARBITRARY_LOG(ros_log_mock).Times(AnyNumber()); // Ensure that arbitrary other logs can occur
```
martiniil, line 66: So what is the point of using `StrictMock<>`?

martiniil (commented):
I am not really aware why you did these changes and why it works. However I have one specific question, see inline comment.

PR #467, opened 2020-11-12
Use github actions

martiniil:
- 0b73a65 can be reverted if we apply https://github.com/PilzDE/industrial_ci_addons/pull/2
- We need to change the settings (branch protection rule) once this PR is merged. (github still expects a travis check, see below)


SansoneG (approved):
Looks good.
I dont quite understand what should be changed in the branch protection rules though. Don't we just need to merge this with administrator rights so github actions is used on the default branch and then everything else should be good to go?

martiniil:
> Looks good.
> I dont quite understand what should be changed in the branch protection rules though. Don't we just need to merge this with administrator rights so github actions is used on the default branch and then everything else should be good to go?

You will see once it is merged and a new build job is started :)

PR #476, opened 2021-01-26
Adapt README.md for noetic

martiniil:
## Description

Add adapted branching model & ci badge

### Things to add, update or check by the maintainers before this PR can be merged.
* [x] Package Readme ([example pilz_robots](https://github.com/PilzDE/pilz_robots/blob/melodic-devel/README.md))
* [x] Good commit messages ([some tips](https://dev.to/jacobherrington/how-to-write-useful-commit-messages-my-commit-message-template-20n9))
* [x] CHANGELOG.rst updated

### Review Checklist
* [x] Documentation describes purpose of file(s) and responsibilities
* [x] Check also #477, #478

PR #477, opened 2021-01-26
Adjust branching model in README.md [kinetic]

martiniil:
Keep branching model updated. See #476


jschleicher (approved):

jschleicher:
For the failing CI check of catkin_lint see #479

